PR #140, opened 2014-08-01
Fix search for nova instance

dirkmueller:
When nova-multi-controller and neutron-server are deployed
on two different nodes, the neutron.conf was not properly updated
with the nova settings because it could't find the node. Search
for the instance (optional since normal deployment order currently
is neutron first, then nova) properly so that get_instance()
in the chef recipe can do its job.

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_config.rb
+++ b/chef/cookbooks/neutron/recipes/common_config.rb
@@ -83,7 +83,15 @@ else
   bind_port = neutron[:neutron][:api][:service_port]
 end
 
-nova = get_instance("roles:nova-multi-controller")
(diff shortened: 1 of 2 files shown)


chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -84,6 +84,15 @@
 end
 
 nova = get_instance("roles:nova-multi-controller")
```
vuntz: Shouldn't we get rid of this line?

nkrinner:
Not tested, but looks good to me. +1

rhafer:
+1

PR #161, opened 2014-10-23
Neutron: Add support for VPNaaS (WIP)

dirkmueller:
This is an optional service plugin, disabled by default


dirkmueller:
lets close this for Cloud 4.x

PR #174, opened 2015-01-17
Remove || nil that makes no sense

vuntz:
The line defining the variable is:
  nova_ssl_cacerts = nova[:nova][:ssl][:ca_certs] unless nova_insecure

nova_ssl_cacerts will be set to nil if nova_insecure is true. And "||
nil" has no impact on empty strings, so it's just redundant.


dirkmueller:
+1

PR #175, opened 2015-01-21
Allow multiple ml2 type drivers at the same time

toabctl:
Restructure the UI to reflect the current situation where users can
choose between ml2 or vmware as core plugin. Selecting ml2 has then
ml2_type_drivers and ml2_mechanism_drivers to configure. For both are
multiple values allowed.


chef/data_bags/crowbar/bc-template-neutron.json
```diff
@@ -67,7 +68,7 @@
     "neutron": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 21,
+      "schema-revision": 30,
```
vuntz: Move to 22, I think 20-30 is for tex.

crowbar_framework/app/views/barclamp/neutron/_edit_attributes.html.haml
```diff
@@ -14,8 +14,11 @@
 
       = select_field :networking_plugin, :collection => :networking_plugins_for_neutron
 
-      #mode_container
-        = select_field :networking_mode, :collection => :networking_modes_for_neutron
+      #ml2_mechanism_drivers_container
+        = string_field %w(ml2_mechanism_drivers)
```
vuntz: Should this really be a string field? Could we have something more magic where people can check the bits they want?

chef/data_bags/crowbar/migrate/neutron/030_seperate_ml2_drivers.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  if a['networking_plugin'] == 'ml2'
+    if a['ml2_mechanism_drivers'].include? 'linuxbridge'
+      a['networking_plugin'] = 'linuxbridge'
+    elsif a['ml2_mechanism_drivers'].include? 'openvswitch'
+      a['networking_plugin'] = 'openvswitch'
+    elsif a['ml2_mechanism_drivers'].include? 'cisco'
+      a['networking_plugin'] = 'cisco'
+    elsif a['ml2_mechanism_drivers'].include? 'cisco'
+      a['networking_plugin'] = 'cisco'
```
vuntz: Duplicated lines.

crowbar_framework/app/views/barclamp/neutron/_edit_attributes.html.haml
toabctl: > Should this really be a string field? Could we have something more magic where people can check the bits they want?

The thing is that the list must be sortable. The order of the elements is important. I had first a multi-select field but I was unable to order the fields. I already asked @tboerger abbout that but sems that there's no easy way to solve it. But I completly agree - it's ugly and doesn't bring any benefit compared to the raw edit mode.

vuntz: I'm mostly worried about two things: people not knowing what they can enter in the field (so making this useless; could be fixed with a help label), and people putting wrong drivers in this field (can be fixed with validation).

toabctl: > I'm mostly worried about two things: people not knowing what they can enter in the field (so making this useless; could be fixed with a help label), and people putting wrong drivers in this field (can be fixed with validation).

How can I add a help label? Validation is already done.

chef/data_bags/crowbar/migrate/neutron/030_seperate_ml2_drivers.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  if a['networking_plugin'] == 'linuxbridge'
+    a['networking_plugin'] = 'ml2'
+    a['ml2_mechanism_drivers'] = 'linuxbridge'
+  elsif a['networking_plugin'] == 'openvswitch'
+    a['networking_plugin'] = 'ml2'
+    a['ml2_mechanism_drivers'] = 'openvswitch'
+  elsif a['networking_plugin'] == 'cisco'
+    a['networking_plugin'] = 'ml2'
+    a['ml2_mechanism_drivers'] = 'cisco'
```
rhafer: The ml2 driver for cisco is called: ```cisco_nexus``` I wonder if we should use that name here as well.

And then there is the problem that this cannot be used standalone, it always needs be be used together with another mech driver, up to now we forced that to be ```openvswitch```. But ```linuxbridge``` might work as well.

chef/data_bags/crowbar/migrate/neutron/030_seperate_ml2_drivers.rb
tboerger: And i think something like this is better readable

```
case
when foo.include? 'a'

when foo.include? 'b'

else

end
```

crowbar_framework/app/assets/javascripts/barclamps/neutron/application.js
```diff
@@ -215,9 +215,10 @@
   };
 
   CiscoPorts.prototype.visualSwitches = function() {
-    return !$.isEmptyObject(this.retrieveSwitches())
-      && $('#networking_plugin').val() == 'cisco'
-      && $('#networking_mode').val() == 'vlan';
+      return !$.isEmptyObject(this.retrieveSwitches())
```
tboerger: Wrong indent, remove 2 spaces.

crowbar_framework/app/assets/javascripts/barclamps/neutron/application.js
```diff
@@ -36,7 +36,7 @@
   CiscoPorts.prototype.initialize = function() {
     var self = this;
 
-    $("#networking_plugin, #networking_mode").live(
+      $("#networking_plugin, #ml2_type_drivers, #ml2_mechanism_drivers").live(
```
tboerger: Wrong indent, remove 2 spaces.

crowbar_framework/app/assets/javascripts/barclamps/neutron/application.js
```diff
@@ -227,73 +228,44 @@
...
-        $('#mode_container').hide(100).attr('disabled', 'disabled');
-
-        $('#vmware_container').hide(100).attr('disabled', 'disabled');
-        break;
-      case 'openvswitch':
-        if (non_forced_mode != undefined) {
-          networking_mode.val(non_forced_mode);
-          networking_mode.removeData('non-forced');
+    $('#networking_plugin').on('change', function() {
+        var plugin = $(this).val();
```
tboerger: Wrong indent, remove 2 spaces, we are writing everything with 2 spaces indent.

tboerger: And this var definition is useless as the variable gets used only once.

tboerger:
Do i see it correct that the cisco functionality have been replaced with the ml2 stuff?

chef/data_bags/crowbar/migrate/neutron/030_seperate_ml2_drivers.rb
toabctl: > Duplicated lines.

Done in https://github.com/toabctl/barclamp-neutron/commit/bb6f0757b4155c9fb84e87254977e84028a66cbc

chef/data_bags/crowbar/migrate/neutron/030_seperate_ml2_drivers.rb
toabctl: > The ml2 driver for cisco is called: cisco_nexus I wonder if we should use that name here as well.

Done in https://github.com/toabctl/barclamp-neutron/commit/bb6f0757b4155c9fb84e87254977e84028a66cbc

crowbar_framework/app/assets/javascripts/barclamps/neutron/application.js
toabctl: > Wrong indent, remove 2 spaces.

All js indent done in https://github.com/toabctl/barclamp-neutron/commit/bb6f0757b4155c9fb84e87254977e84028a66cbc

crowbar_framework/app/assets/javascripts/barclamps/neutron/application.js
toabctl: > And this var definition is useless as the variable gets used only once.

Done in https://github.com/toabctl/barclamp-neutron/commit/bb6f0757b4155c9fb84e87254977e84028a66cbc

crowbar_framework/app/views/barclamp/neutron/_edit_attributes.html.haml
tboerger: Example: https://github.com/crowbar/barclamp-dns/blob/master/crowbar_framework/app/views/barclamp/dns/_edit_attributes.html.haml#L17-L18

crowbar_framework/app/assets/javascripts/barclamps/neutron/application.js
```diff
@@ -1,6 +1,6 @@
 /**
  * Copyright 2011-2013, Dell
- * Copyright 2013-2014, SUSE LINUX Products GmbH
+ * Copyright 2013-2015, SUSE LINUX Products GmbH
```
dirkmueller: I hope noone tells @tboerger that we're "SUSE Linux GmbH" now..

tboerger: You are so funny :)

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -94,14 +94,31 @@ def validate_proposal_after_save proposal
...
 
     plugin = proposal["attributes"]["neutron"]["networking_plugin"]
-    mode = proposal["attributes"]["neutron"]["networking_mode"]
+    ml2_mechanism_drivers = proposal["attributes"]["neutron"]["ml2_mechanism_drivers"]
+    ml2_type_drivers = proposal["attributes"]["neutron"]["ml2_type_drivers"]
 
-    if %(linuxbridge cisco).include?(plugin) and mode != "vlan"
-      validation_error("The \"#{plugin}\" plugin only supports the mode: \"vlan\"")
+    # validate given comma seperated driver list
+    ml2_mechanism_drivers_valid = %w(linuxbridge openvswitch cisco_nexus")
```
dirkmueller: maybe do that as an attribute ?

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -94,14 +94,31 @@ def validate_proposal_after_save proposal
...
+    ml2_type_drivers_proposed = ml2_type_drivers.split(",")
+    ml2_type_drivers_proposed.each do |drv|
+      unless ml2_type_drivers_valid.include? drv.strip
+        validation_error("ML2 type driver '#{drv}' not allowed. Valid drivers are: #{ml2_type_drivers_valid.join(', ')}")
+      end
     end
 
-    if !%(gre vlan).include?(mode)
-      validation_error("Unknown networking mode \"#{mode}\"")
+    %w(linuxbridge cisco_nexus).each do |type_drv|
```
dirkmueller: suggest to rename |type_drv| to |drv| since its a mechanism driver, not a type driver

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -94,14 +94,31 @@ def validate_proposal_after_save proposal
...
+      if ml2_mechanism_drivers_proposed.include? type_drv and not ml2_type_drivers_proposed.include? 'vlan'
+        validation_error("The type driver \"#{type_drv}\" only supports the mechanism driver \"vlan\"")
```
dirkmueller: I think 'type' and 'mechanism' needs to be exchanged here.

crowbar_framework/app/views/barclamp/neutron/_edit_attributes.html.haml
rhafer: > The thing is that the list must be sortable.

Is that really true for the mech_driver? I thought that would only be the case for the type drivers but ICBW. If it's not needed to be sortable we could make it a group of checkboxes, couldn't we?

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -51,19 +51,21 @@
   subscribes :run, resources(:cookbook_file=> disable_rp_filter_file), :delayed
 end
 
-
 case neutron[:neutron][:networking_plugin]
```
rhafer: How about turning this into an ```if neutron[:neutron][:networking_plugin] == "ml2"```? Currently ```ml2``` is really the only ```when``` branch here as e.g. ```vmware``` doesn't have any l2 agent.

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -108,9 +110,14 @@
       group node[:neutron][:platform][:group]
     end
   end
+when "vmware"
```
rhafer: why?

crowbar_framework/app/models/neutron_service.rb
toabctl: > suggest to rename |type_drv| to |drv| since its a mechanism driver, not a type driver

Done

crowbar_framework/app/models/neutron_service.rb
toabctl: > I think 'type' and 'mechanism' needs to be exchanged here.

Done

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -185,56 +180,61 @@
 num_vlans = neutron[:neutron][:num_vlans]
 vlan_end = [vlan_start + num_vlans - 1, 4094].min
 
+
 case neutron[:neutron][:networking_plugin]
```
rhafer: Just thinking out loud (so this should not block this PR, just some idea for cleanup): If I am not mistaken, everything from here down to the ```end``` is either related to the l2 agent configuration or to the configuration of the nsx plugin. Wouldn't it make sense to move the l2-agent related bits to ```common_agent``` and the nsx plugin related things to ```server.rb```?

That way the bits would only get executed where they are really needed.

rhafer: BTW, if you decide to do that cleanup, please do it as a separate commit :)

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -185,56 +180,61 @@
...
+      cookbook "neutron"
+      source "ovs_neutron_plugin.ini.erb"
+      owner "root"
+      group neutron[:neutron][:platform][:group]
+      mode "0640"
+      variables(
+        :physnet => neutron[:neutron][:ml2_type_drivers].include?('gre') ? nil : "br-fixed",
+        :networking_mode => neutron[:neutron][:networking_mode],
+        :vlan_start => vlan_start,
+        :vlan_end => vlan_end
```
rhafer: Just for completeness (we talked about that already). Everything else but ```physnet``` should be no longer required here (they are all options for the plugin not for the agent. Do you plan to address that in this PR? If yes, I'd like to see it as a separate commit.

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -185,56 +180,61 @@
...
+    template agent_config_path do
+      cookbook "neutron"
+      source "linuxbridge_conf.ini.erb"
+      owner "root"
+      group neutron[:neutron][:platform][:group]
+      mode "0640"
+      variables(
+        :physnet => (node[:crowbar_wall][:network][:nets][:nova_fixed].first rescue nil),
+        :vlan_start => vlan_start,
+        :vlan_end => vlan_end
```
rhafer: Same here. We only need the ```physnet``` parameter

chef/cookbooks/neutron/recipes/post_install_conf.rb
```diff
@@ -61,26 +62,35 @@ def mask_to_bits(mask)
...
 
+fixed_network_type = ""
 floating_network_type = ""
-if node[:neutron][:networking_mode] == 'vlan'
-  fixed_network_type = "--provider:network_type vlan --provider:segmentation_id #{fixed_net["vlan"]} --provider:physical_network physnet1"
-  if node[:network][:networks][:nova_floating][:use_vlan]
-    floating_network_type = "--provider:network_type vlan --provider:segmentation_id #{floating_net["vlan"]} --provider:physical_network physnet1"
-  else
+
+# TODO(toabctl): For now we just use the first ml2_type_driver for foating network type
```
rhafer: "foating"

rhafer:
Btw, you're breaking the cisco_nexus support in: https://github.com/crowbar/barclamp-neutron/pull/175/files#diff-c147bad02dfed1864bcb98c50ffd226fR173
You'd need to check for ```ml2_mech_drivers.include?("cisco") there.

toabctl:
> Btw, you're breaking the cisco_nexus support in: https://github.com/crowbar/barclamp-neutron/pull/175/files#diff-c147bad02dfed1864bcb98c50ffd226fR173
> You'd need to check for ```ml2_mech_drivers.include?("cisco") there.

Done in https://github.com/crowbar/barclamp-neutron/commit/6573fd00800313318448834fb6177af75676afa8 . And it must be "cisco_nexus".

chef/cookbooks/neutron/recipes/post_install_conf.rb
toabctl: > "foating"

Done in https://github.com/crowbar/barclamp-neutron/commit/6573fd00800313318448834fb6177af75676afa8

chef/cookbooks/neutron/recipes/common_config.rb
toabctl: I would like to fixed the cleanup things when we are done with this PR. The change is already huge...

chef/cookbooks/neutron/recipes/common_config.rb
toabctl: I would like to fixed the cleanup things when we are done with this PR. The change is already huge...

chef/cookbooks/neutron/recipes/common_agent.rb
toabctl: > why?

Replaced with an ```if``` in https://github.com/crowbar/barclamp-neutron/commit/6573fd00800313318448834fb6177af75676afa8

chef/cookbooks/neutron/recipes/common_agent.rb
toabctl: > How about turning this into an if neutron[:neutron][:networking_plugin] == "ml2"? Currently ml2 is really the only when branch here as e.g. vmware doesn't have any l2 agent.

Done in https://github.com/crowbar/barclamp-neutron/commit/6573fd00800313318448834fb6177af75676afa8

chef/cookbooks/neutron/attributes/default.rb
```diff
@@ -20,8 +20,7 @@
 default[:neutron][:debug] = false
 default[:neutron][:verbose] = false
 default[:neutron][:dhcp_domain] = "openstack.local"
-default[:neutron][:networking_mode] = "gre"
-default[:neutron][:networking_plugin] = "openvswitch"
+default[:neutron][:networking_plugin] = "ml2"
 default[:neutron][:cisco_support] = false
```
vuntz: Not related to your PR, but can you drop this attribute? I don't think it's used.

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -108,7 +109,11 @@
   end
 end
 
-if ['openvswitch', 'cisco', 'vmware'].include? neutron[:neutron][:networking_plugin]
+
+# openvswitch installation and configuration
+if neutron[:neutron][:networking_plugin] == 'vmware' or
+  (neutron[:neutron][:networking_plugin] == 'ml2' and
+   neutron[:neutron][:ml2_mechanism_drivers].include?("openvswitch"))
```
vuntz: The old test was mentioning cisco, this seems to be lost here?

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -49,19 +49,20 @@
...
+  case
+  when ml2_mech_drivers.include?("openvswitch")
+    neutron_agent = node[:neutron][:platform][:ovs_agent_name]
+    neutron_agent_pkg = node[:neutron][:platform][:ovs_agent_pkg]
+    agent_config_path = "/etc/neutron/plugins/openvswitch/ovs_neutron_plugin.ini"
+  when ml2_mech_drivers.include?("linuxbridge")
+    neutron_agent = node[:neutron][:platform][:lb_agent_name]
+    neutron_agent_pkg = node[:neutron][:platform][:lb_agent_pkg]
+    agent_config_path = "/etc/neutron/plugins/linuxbridge/linuxbridge_conf.ini"
+  end
```
vuntz: So if we specify `mech_drivers = ["openvswitch", "linuxbridge"]`, when will the linuxbridge agent ever be installed?

chef/cookbooks/neutron/recipes/common_agent.rb
toabctl: > The old test was mentioning cisco, this seems to be lost here?

When selecting cisco, openvswitch is mandatory. So checking for openvswitch is enough.

chef/cookbooks/neutron/recipes/common_agent.rb
toabctl: > So if we specify mech_drivers = ["openvswitch", "linuxbridge"], when will the linuxbridge agent ever be installed?

You can't use both in parallel. See https://github.com/toabctl/barclamp-neutron/commit/a66c76cac342de85055734064d93c86b1eff5542#diff-1cab7f3b6842279ae4d6bd875e073cacR114

chef/data_bags/crowbar/migrate/neutron/022_seperate_ml2_drivers.rb
```diff
@@ -0,0 +1,57 @@
+def upgrade ta, td, a, d
+  if a['networking_mode']
+    a['ml2_type_drivers'] = a['networking_mode']
+  else
+    a['ml2_type_drivers'] = ta['ml2_type_drivers']
+  end
+
+  if a['networking_plugin'] == 'linuxbridge'
+    a['networking_plugin'] = 'ml2'
+    a['ml2_mechanism_drivers'] = 'linuxbridge'
```
rhafer: ```ml2_mechanism_drivers``` is supposed to be a list, so shouldn't this be ```['linuxbridge']```

chef/data_bags/crowbar/migrate/neutron/022_seperate_ml2_drivers.rb
```diff
@@ -0,0 +1,57 @@
+def upgrade ta, td, a, d
+  if a['networking_mode']
+    a['ml2_type_drivers'] = a['networking_mode']
```
rhafer: ```ml2_type_drivers``` is a list. So I guess this needs to be ```a['ml2_type_drivers'] = [a['networking_mode']] ```

chef/data_bags/crowbar/migrate/neutron/022_seperate_ml2_drivers.rb
```diff
@@ -0,0 +1,57 @@
...
+  else
+    a['ml2_type_drivers'] = ta['ml2_type_drivers']
+  end
+
+  if a['networking_plugin'] == 'linuxbridge'
+    a['networking_plugin'] = 'ml2'
+    a['ml2_mechanism_drivers'] = 'linuxbridge'
+  elsif a['networking_plugin'] == 'openvswitch'
+    a['networking_plugin'] = 'ml2'
+    a['ml2_mechanism_drivers'] = 'openvswitch'
```
rhafer: Same here.

chef/data_bags/crowbar/migrate/neutron/022_seperate_ml2_drivers.rb
```diff
@@ -0,0 +1,57 @@
...
+  elsif a['networking_plugin'] == 'cisco'
+    a['networking_plugin'] = 'ml2'
+    a['ml2_mechanism_drivers'] = 'cisco_nexus'
```
rhafer: I think we need to add openvswitch to the list here as well, don't we?

chef/data_bags/crowbar/migrate/neutron/022_seperate_ml2_drivers.rb
```diff
@@ -0,0 +1,57 @@
...
+  elsif a['networking_plugin'] == 'vmware'
+    a['networking_plugin'] = 'vmware'
+  else
+    a['networking_plugin'] = ta['networking_plugin']
+    a['ml2_mechanism_drivers'] = ta['ml2_mechanism_drivers']
```
rhafer: Is this a scenario that really can happen?

crowbar_framework/app/helpers/barclamp/neutron_helper.rb
```diff
@@ -20,22 +20,33 @@ module NeutronHelper
...
     end
 
-    def networking_modes_for_neutron(selected)
+    def networking_ml2_mechanism_drivers_for_neutron(selected)
+      selected = selected.gsub(/\s+/, "").split(",")
       options_for_select(
         [
-          ["gre", "gre"], 
-          ["vlan", "vlan"]
+          ["linuxbridge", "linuxbridge"],
```
bkutil: Sorting here?

chef/data_bags/crowbar/migrate/neutron/022_seperate_ml2_drivers.rb
toabctl: > Is this a scenario that really can happen?

I don't think so. I can remove it if you want.

chef/data_bags/crowbar/migrate/neutron/022_seperate_ml2_drivers.rb
toabctl: > ml2_mechanism_drivers is supposed to be a list, so shouldn't this be ['linuxbridge']

Done in https://github.com/toabctl/barclamp-neutron/commit/e031434a8db081f91d103898fb8c5486eaf69621

chef/data_bags/crowbar/migrate/neutron/022_seperate_ml2_drivers.rb
toabctl: > ml2_type_drivers is a list. So I guess this needs to be a['ml2_type_drivers'] = [a['networking_mode']]

Done in https://github.com/toabctl/barclamp-neutron/commit/e031434a8db081f91d103898fb8c5486eaf69621

chef/data_bags/crowbar/migrate/neutron/022_seperate_ml2_drivers.rb
toabctl: > Same here.

Done in https://github.com/toabctl/barclamp-neutron/commit/e031434a8db081f91d103898fb8c5486eaf69621

chef/data_bags/crowbar/migrate/neutron/022_seperate_ml2_drivers.rb
toabctl: > I think we need to add openvswitch to the list here as well, don't we?

Correct. Done in https://github.com/toabctl/barclamp-neutron/commit/e031434a8db081f91d103898fb8c5486eaf69621

rhafer:
I gave this some testing (together with https://github.com/crowbar/barclamp-nova/pull/419) in HA and non-HA setup. This looks pretty good to me now. (Creating and using provider networks of different types worked just fine)

+1

chef/cookbooks/neutron/recipes/common_agent.rb
vuntz: One more question: I don't see any validation at https://github.com/crowbar/barclamp-neutron/pull/175/files#diff-1cab7f3b6842279ae4d6bd875e073cacR97 of the fact that ovs or linuxbridge must be specified. Is this not mandatory to have one of the two? If not, then where are we defining the variables above in that third case?

chef/cookbooks/neutron/recipes/cisco_support.rb
```diff
@@ -2,7 +2,7 @@
 
 neutron = node
 
-if neutron[:neutron][:networking_mode] == 'vlan'
+if neutron[:neutron][:ml2_type_drivers].include? 'vlan'
```
vuntz: Not that it hurts (on the contrary, I'd leave it), and more for my curiosity; aren't we enforcing through the validation the use of vlan when cisco is used?

chef/cookbooks/neutron/templates/default/linuxbridge_conf.ini.erb
```diff
@@ -8,7 +8,8 @@
 #
 # tenant_network_type = local
 # Example: tenant_network_type = vlan
-tenant_network_type = vlan
+# NOTE(rhafer): tenant_network_type is longer used for ml2. The option is now
```
vuntz: s/longer/no longer/

chef/cookbooks/neutron/templates/default/linuxbridge_conf.ini.erb
```diff
@@ -20,7 +21,8 @@ tenant_network_type = vlan
 #
 # network_vlan_ranges =
 # Example: network_vlan_ranges = physnet1:1000:2999
-network_vlan_ranges = physnet1:<%= @vlan_start %>:<%= @vlan_end %>
+# NOTE(rhafer): network_vlan_ranges is longer used for ml2. The option is now
```
vuntz: same here

chef/cookbooks/neutron/recipes/cisco_support.rb
toabctl: > Not that it hurts (on the contrary, I'd leave it), and more for my curiosity; aren't we enforcing through the validation the use of vlan when cisco is used?

Yes, we do.

chef/cookbooks/neutron/templates/default/linuxbridge_conf.ini.erb
toabctl: > s/longer/no longer/

Fixed in https://github.com/toabctl/barclamp-neutron/commit/fc213eb8071c32aefef5ed4e833db517db9c5f8f

chef/cookbooks/neutron/templates/default/linuxbridge_conf.ini.erb
toabctl: > same here

Fixed in https://github.com/toabctl/barclamp-neutron/commit/fc213eb8071c32aefef5ed4e833db517db9c5f8f

chef/cookbooks/neutron/recipes/common_agent.rb
toabctl: > One more question: I don't see any validation at https://github.com/crowbar/barclamp-neutron/pull/175/files#diff-1cab7f3b6842279ae4d6bd875e073cacR97 of the fact that ovs or linuxbridge must be specified. Is this not mandatory to have one of the two? If not, then where are we defining the variables above in that third case?

It's not allowed to set no mechanism driver (due to the 'required' flag from https://github.com/crowbar/barclamp-neutron/pull/175/files#diff-c0cfbff55458d6b05ced40f360e9ab75R20 ). When you select 'cisco_nexus' there's a check that 'openvswitch ' must be selected as well. So it's not possible to have non of 'openvswitch/linuxbridge' selected.

crowbar_framework/config/locales/neutron/en.yml
```diff
@@ -33,7 +33,12 @@ en:
         dhcp_domain: 'DHCP Domain'
         plugin_header: 'Networking Plugin'
         networking_plugin: 'Plugin'
-        networking_mode: 'Mode'
+        ml2_type_drivers: 'Modular Layer 2 type drivers'
+        ml2_type_drivers_default_provider_network: 'Default type driver for provider network'
+        ml2_type_drivers_default_provider_network_hint: 'Modular Layer 2 type driver used to create the "fixed" network'
```
vuntz: Can you put a list of valid values here?

crowbar_framework/config/locales/neutron/en.yml
```diff
@@ -33,7 +33,12 @@ en:
...
+        ml2_type_drivers_default_tenant_network: 'Default type driver for tenant network'
+        ml2_type_drivers_default_tenant_network_hint: 'Modular Layer 2 type driver used to create the tenant networks'
```
vuntz: Same thing, list of valid values would help.

vuntz: Oh, sorry, wrong string. Ignore these comments, will add new ones :-)

chef/cookbooks/neutron/recipes/l3_ha.rb
```diff
@@ -65,18 +65,26 @@
...
+  when ml2_mech_drivers.include?("openvswitch")
+    neutron_agent = node[:neutron][:platform][:ovs_agent_name]
+    neutron_agent_ra = node[:neutron][:ha][:l3]["openvswitch_ra"]
+  when ml2_mech_drivers.include?("linuxbridge")
+    neutron_agent = node[:neutron][:platform][:lb_agent_name]
+    neutron_agent_ra = node[:neutron][:ha][:l3]["linuxbridge_ra"]
+  end
 when "vmware"
   neutron_agent = ''
+  neutron_agent_ra = node[:neutron][:ha][:l3]["vmware_ra"]
```
vuntz: This should be an empty string, as there's no l3 service with vmware (and this attribute doesn't exist)

chef/cookbooks/neutron/recipes/post_install_conf.rb
```diff
@@ -61,26 +62,33 @@ def mask_to_bits(mask)
...
-  else
+
+networking_plugin = node[:neutron][:networking_plugin]
+ml2_type_drivers_default_provider_network = node[:neutron][:ml2_type_drivers_default_provider_network]
+case networking_plugin
+when 'ml2'
+  case ml2_type_drivers_default_provider_network
+  when 'vlan'
+    fixed_network_type = "--provider:network_type vlan --provider:segmentation_id #{fixed_net["vlan"]} --provider:physical_network physnet1"
     floating_network_type = "--provider:network_type flat --provider:physical_network physnet1"
```
vuntz: Hrm.  Shouldn't we pass `--provider:segmentation_id #{floating_net["vlan"]}` ?

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -149,26 +149,26 @@
...
 directory "/etc/neutron/plugins/ml2" do
   mode 0755
   action :create
   only_if { node[:platform] == "ubuntu" }
 end
 
+# NOTE(toabctl): tenant_network types should have as first element 'ml2_type_drivers_default_tenant_network' and then the rest of the selected type drivers.
+#                when creating tenant networks, it's not possible to manually select the network type. Neutron just tries every selected type until success
+#                so the order is important
+tenant_network_types = [node[:neutron][:ml2_type_drivers_default_tenant_network]] + node[:neutron][:ml2_type_drivers].collect{ |el| el if el != node[:neutron][:ml2_type_drivers_default_tenant_network]}.compact
```
vuntz: Can't we write this as:
```

tenant_network_types = [[node[:neutron][:ml2_type_drivers_default_tenant_network]] + node[:neutron][:ml2_type_drivers]].flatten.uniq
```

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -94,14 +94,40 @@ def validate_proposal_after_save proposal
...
+    # cisco_nexus mech driver needs also openvswitch mech driver
+    # TODO(toabctl): select openvswitch automatically if cisco_nexus was selected!?
+    if ml2_mechanism_drivers.include? "cisco_nexus" and not ml2_mechanism_drivers.include? "openvswitch"
+      validation_error("The 'cisco_nexus' mechanism driver needs also the 'openvswitch' mechanism driver")
+    end
+
+    # for now, openvswitch and linuxbrige can't be used in parallel
+    if ml2_mechanism_drivers.include? "openvswitch" and ml2_mechanism_drivers.include? "linuxbridge"
+      validation_error("The 'openvswitch' and 'linuxbridge' mechanism drivers can't be used in parallel. Only select one of them")
     end
```
vuntz: Shouldn't we validate that `ml2_mechanism_drivers` should have at least one of openvswitch or linuxbridge?

Similarly, for ml2_type_drivers, shouldn't we check it has at least vlan or gre?

(can be done in a later PR if tracked)

toabctl: > Shouldn't we validate that ml2_mechanism_drivers should have at least one of openvswitch or linuxbridge?
> 
> Similarly, for ml2_type_drivers, shouldn't we check it has at least vlan or gre?

That's already checked because we set 'required' to True in the barclamp schema.

chef/cookbooks/neutron/recipes/post_install_conf.rb
rhafer: I don't think it makes a lot of sense to do that with  ```--provider:network_type flat```

But actually there might be another issue here. We lost the:
```
if node[:network][:networks][:nova_floating][:use_vlan]
    floating_network_type = "......."
```
thing.

crowbar_framework/app/models/neutron_service.rb
vuntz: No, that means that there will be a list. It won't validate that the content of the list is valid.

toabctl: Right. I added a check in https://github.com/crowbar/barclamp-neutron/commit/1504fd042f9cf6bf1e7665d354e3e84e141adf39 that there is at least 1 mech/type driver selected when ml2 is used

chef/cookbooks/neutron/recipes/server.rb
toabctl: Done in https://github.com/crowbar/barclamp-neutron/commit/1504fd042f9cf6bf1e7665d354e3e84e141adf39

chef/cookbooks/neutron/recipes/l3_ha.rb
toabctl: Done in https://github.com/crowbar/barclamp-neutron/commit/1504fd042f9cf6bf1e7665d354e3e84e141adf39

chef/cookbooks/neutron/recipes/post_install_conf.rb
toabctl: > But actually there might be another issue here. We lost the:
> 
> if node[:network][:networks][:nova_floating][:use_vlan]
>     floating_network_type = "......."
> thing. Which makes we wonder why floating IPs still worked yesterday ... I'll recheck.

As discussed with @rhafer I readded the current behavior in https://github.com/crowbar/barclamp-neutron/commit/ccf35be763d9761c70c757f1b8d9789de4630d03

rhafer: > As discussed with @rhafer I readded the current behavior in ccf35be

Just for completeness: The missing "--provider:network_type vlan" indeed broke floating IPs when using the ```linuxbridge``` mechanism driver. (It didn't do any harm for the openvswithc driver that's why I didn't notice earlier)

vuntz:
Let's check it in.

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -1,5 +1,5 @@
 # Copyright 2013 Dell, Inc.
-# Copyright 2014 SUSE
+# Copyright 2014-2015 SUSE Linux GmbH
```
tboerger: That is something that have to be replaced everywhere :(

PR #178, opened 2015-01-22
Use sysconfig file to figure out the db-migrate arguments

rhafer:
Is avoids having to update the paths in two places when changes are required.

### DIFF
--- a/chef/cookbooks/neutron/recipes/server.rb
+++ b/chef/cookbooks/neutron/recipes/server.rb
@@ -137,6 +137,16 @@ template "/etc/sysconfig/neutron" do
   notifies :restart, "service[#{node[:neutron][:platform][:service_name]}]"
 end
 
+template "/etc/default/neutron-server" do
+  source "neutron-server.erb"
+  owner "root"
+  group node[:neutron][:platform][:group]
+  variables(
+      :neutron_plugin_config => "/etc/neutron/plugins/ml2/ml2_conf.ini"
+    )
+  only_if { node[:platform] == "ubuntu" }
+end
+
 directory "/var/cache/neutron" do
   owner node[:neutron][:user]
   group node[:neutron][:group]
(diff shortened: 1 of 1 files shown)


vuntz:
+1

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -184,15 +182,14 @@
...
-config_files = "--config-file /etc/neutron/neutron.conf --config-file #{plugin_cfg_path}"
-if node[:neutron][:networking_plugin] == "cisco"
-  config_files += " --config-file /etc/neutron/plugins/ml2/ml2_conf_cisco.ini"
-end
-
 execute "neutron-db-manage migrate" do
   user node[:neutron][:user]
   group node[:neutron][:group]
-  command "neutron-db-manage #{config_files} upgrade head"
+  command 'source /etc/sysconfig/neutron; \
```
dirkmueller: the writing of that file is in a "only_if suse" clause. so this needs to be added here as well imho..

do we need an explicit "notify" of this execute block from the template block that writes the sysconfig file? I'm unsure.

rhafer: > the writing of that file is in a "only_if suse" clause. so this needs to be added here as well imho..

Good point. I think it's only needed on SUSE anyway.

> do we need an explicit "notify" of this execute block from the template block that writes the sysconfig file? I'm unsure.

I don't think so. We want the migration to run only once. Triggering it on sysconfig change wouldn't also do anything normally as that would not change the "head" of the migrations queue. (AFAIK with Juno this is also less of an issue compared to Icehouse as IIRC nowadays the migrations in neutron are mostly independent of the actual plugin configuration)

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -184,18 +182,18 @@
...
+           for i in $NEUTRON_PLUGIN_CONF; do \
+             CONF_ARGS="$CONF_ARGS --config-file $i"; \
+           done; \
+           neutron-db-manage --config-file /etc/neutron/neutron.conf $CONF_ARGS upgrade head'
   # We only do the sync the first time, and only if we're not doing HA or if we
   # are the founder of the HA cluster (so that it's really only done once).
-  only_if { !node[:neutron][:db_synced] && (!ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node)) }
+  only_if { node[:platform] == "suse" && !node[:neutron][:db_synced] &&
+            (!ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node)) }
```
vuntz: Hrm, @dirkmueller was right, but the new code is not correct: we still need the db-manager call on non-SUSE. I suggest changing the command depending on the platform instead.

rhafer: I guess than we can just forget about this change :). Because it would re-introduce the duplication. Well I guess I could also rework it using a slightly different approach. But can you explain to me (again) why we need this code on non-SUSE (so non-HA) deployments?

vuntz: Well, because we shouldn't assume that non-SUSE implies non-HA :-)

vuntz:
@rhafer: I sent https://github.com/rhafer/barclamp-neutron/pull/1 to address my concern. This will still need a rebase, though.

rhafer:
@vuntz thanks, but that patch doesn't really fix it. I'll take another look into as soon as possible.

rhafer:
Rebased and merged @vuntz' additions

vuntz:
+1, but it doesn't really count since I worked on it :-)

PR #181, opened 2015-01-29
Make sure to not write an empty database setting to neutron.conf

rhafer:
This confuses the metadata agent and causes it to fail on startup.


toabctl:
+1

vuntz:
In which case can it be empty?

rhafer:
> In which case can it be empty?

 ```sql_connection``` is only stored as a node attribute on the node with the neutron-server role (see https://github.com/crowbar/barclamp-neutron/blob/master/chef/cookbooks/neutron/recipes/database.rb#L64). So when you deploy the neutron-l3 role to a node other that neutron-server the ```node``` object doesn't have that attribute (and it doesn't need it).

vuntz:
Thanks, +1

PR #183, opened 2015-01-30
Disable multiple_api workers when core_plugin vmware (bnc#915572)

mjura:
If you run with multiple_api workers there is a race condition in the nsx sync backend as we start reading from the database before it's connected if running with 1 worker this doesn't occur as things don't occur concurrently.

https://bugs.launchpad.net/neutron/+bug/1354226
and
https://bugzilla.suse.com/show_bug.cgi?id=915572

### DIFF
--- a/chef/cookbooks/neutron/templates/default/neutron.conf.erb
+++ b/chef/cookbooks/neutron/templates/default/neutron.conf.erb
@@ -213,7 +213,11 @@ allow_overlapping_ips = False
 # worker thread in the current process.  Greater than 0 launches that number of
 # child processes as workers.  The parent process manages them.
 # api_workers = 0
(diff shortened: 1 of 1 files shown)


chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -213,7 +213,11 @@ allow_overlapping_ips = False
 # worker thread in the current process.  Greater than 0 launches that number of
 # child processes as workers.  The parent process manages them.
 # api_workers = 0
+# This is workaround for bug1354226
```
dirkmueller: lp#1354226

dirkmueller:
small nitpick, LGTM overall

PR #187, opened 2015-02-03
Added asset symlinking and cleanup

tboerger:


mjura:
I have no idea what it does but indentations looks awesome :)

mjura:
Travis failed, it is sad :(

tboerger:
@mjura Yeah, but it's only failing because the 1.8 test case is still active. that have to be disabled as we don't need the 1.8 tests anymore.

tboerger:
@mjura I have added a seperate commit for removing the unsupported ruby versions from travis testing.

PR #189, opened 2015-02-04
Preselect and hide ml2 driver elements as needed

rhafer:


vuntz:
One thing that is "breaking" from a UX perspective is that if have ovs/gre selected, then pick linuxbridge and pick ovs again, I lose the gre bit as everything is moved to vlan.

I'm wondering if we want some magic to remember what the ovs bits were.

vuntz:
Oh, two other bugs:

- if I pick cisco, then vlan should be selected two in the drivers

- on initial load, if I have ovs with just one driver, I can see the "default type driver" fields. If I then move from vlan to gre (or vice-versa), these fields disappear. And only appear when I select both vlan and gre

crowbar_framework/app/assets/javascripts/barclamps/neutron/application.js
```diff
@@ -253,26 +253,52 @@ function networking_plugin_check() {
...
 function ml2_type_drivers_check() {
-  var values = $('#ml2_type_drivers').val();
+  var values = $('#ml2_type_drivers').val() || [];
   if (values.indexOf("vlan") >= 0) {
     $('#num_vlans_container').show();
   } else {
     $('#num_vlans_container').hide();
   }
+
+  if (values.length == 1) {
```
toabctl: This logic doesn't work. When doing the following, I get ```The default provider network type driver "vlan" is not a selected ml2 type driver ``` as error.
* select both ml2 type drivers
* select for both default values the second driver from the ```ml2 type drivers``` select field
* now select only the first driver in the ```ml2 type drivers``` select field
* press save - here's the error message

rhafer: > This logic doesn't work. When doing the following, I get The default provider network type driver "vlan" is not a selected ml2 type driver as error.

This is weird. I can see that behaviour as well. But it doesn't seem to be an issue with this change. At least I can see the values of the combo boxes correctly updated to only remaining selected type_driver. I'll investigate further.

> Another problem is, that the default ml2 type driver boxes are visible after pressing Apply/Save (and no error occurs) even if only one type driver is availalbe.

This is what @vuntz  already mentioned. Will be fixed with the next iteration.

rhafer: This was just a few missing ```.trigger('change')``` calls in the end.

toabctl:
Works for me now. +1

crowbar_framework/app/assets/javascripts/barclamps/neutron/application.js
vuntz: Should we do `<= 1` instead of `== 1`? (not sure, just asking)

rhafer: Makes sense I guess.

vuntz:
Works nicely.

vuntz:
Just saw @dirkmueller and found a bug: select ml2+ovs+gre; move to ml2+lb; move to vmware; move to ml2 (lb is automatically selected); move to ml2+ovs. You end up with vlan instead of gre.

rhafer:
> Just saw @dirkmueller and found a bug: select ml2+ovs+gre; move to ml2+lb; move to vmware; move to ml2 (lb is automatically selected); move to ml2+ovs. You end up with vlan instead of gre.

Should be fixed now.

PR #190, opened 2015-02-05
Cleanup common_config recipe

rhafer:
Only create the layer 2 agent specific config files where they are needed
(i.e. compute-nodes and l3 node -> common_agent.rb). The vmware specific bits
are only really needed on the neutron server.


rhafer:
This is basically addressing what I raised in: 
https://github.com/crowbar/barclamp-neutron/pull/175#discussion_r23386981

vuntz:
+1

PR #191, opened 2015-02-06
Fix cisco nexus support

rhafer:
We overlooked these two instances when enabling support for multiple ml2 type
drivers.


toabctl:
+1

vuntz:
+1

dirkmueller:
+1

PR #193, opened 2015-02-10
Do not set nova_ca_certificates_file

vuntz:
Everywhere else, we rely on the admins to either properly install their
CA certificate system-wide, or to use the insecure options in the
barclamps.


rhafer:
+1

jsuchome:
The whole configuration is a bit confusing, but yes, this makes it better.

+1

PR #194, opened 2015-02-12
Restart pacemaker when needed on the l3-node (bnc#915176)

rhafer:
When the admin network and the "public" or "nova_fixed" networks are using the
same physical interface, we need to take care to stop the pacemaker services
before enslaving the interface into the newly created openvswitch bridge and to
cleanly restarted afterwards.

https://bugzilla.suse.com/show_bug.cgi?id=915176

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_agent.rb
+++ b/chef/cookbooks/neutron/recipes/common_agent.rb
@@ -152,6 +152,7 @@ if neutron[:neutron][:networking_plugin] == "ml2"
   end
 end
 
+neutron_l3_ha = node.roles.include?("neutron-l3") && node[:neutron][:ha][:l3][:enabled]
 
 # openvswitch installation and configuration
 if neutron[:neutron][:networking_plugin] == 'vmware' or
@@ -245,10 +246,46 @@ if neutron[:neutron][:networking_plugin] == 'vmware' or
       command "ovs-vsctl add-br #{name}; ip link set #{name} up"
       not_if "ovs-vsctl list-br |grep -q #{name}"
     end
-    execute "Neutron: add #{bound_if} to #{name}" do
-      command "ovs-vsctl del-port #{name} #{bound_if} ; ovs-vsctl add-port #{name} #{bound_if}"
-      not_if "ovs-dpctl show system@#{name} | grep -q #{bound_if}"
+
+    # If:
+    #   * this network is on the same physical interface as the admin network
+    #   * we're running in HA mode
+    #   * corosync is configured to use multicast networking
+    # We need to stop openais/corosync before moving the network interface
+    # into the openvswitch. Otherwise we'll break corosync's multicast
+    # communication and cause fencing.
+    # For details: https://bugzilla.suse.com/show_bug.cgi?id=915176
+    handle_corosync_restart = node[:crowbar_wall][:network][:nets]["admin"].last == bound_if &&
+        neutron_l3_ha && node[:corosync][:transport] == "udp"
+
+    not_if_clause = "ovs-vsctl iface-to-br #{bound_if} | grep -q #{name}"
+
+    if handle_corosync_restart
+      # Put the whole cluster into maintenance mode before shutting down
+      # corosync. It's not enough to put only single nodes into maintenance
+      # mode here as the remaining nodes will immediately fence the node in
+      # maintenance once corosync gets stopped on it. (For details see:
+      # https://bugzilla.suse.com/show_bug.cgi?id=917593)
+      #
+      # As this operation affects all cluster nodes only do it on the founder.
+      if (CrowbarPacemakerHelper.is_cluster_founder?(node))
+        execute "enable maintenance mode" do
+          command "crm --wait configure property maintenance-mode=true"
+          not_if not_if_clause
+        end
+      end
+
+      crowbar_pacemaker_sync_mark "sync-neutron-l3_before_network_reconfig" do
+        not_if not_if_clause
+      end
+
+      service "stop #{node[:corosync][:platform][:service_name]} for network reconfiguration" do
+        service_name node[:corosync][:platform][:service_name]
+        action :stop
+        not_if not_if_clause
+      end
     end
+
     ruby_block "Have #{name} usurp config from #{bound_if}" do
       block do
         target = ::Nic.new(name)
@@ -257,10 +294,40 @@ if neutron[:neutron][:networking_plugin] == 'vmware' or
         Chef::Log.info("#{name} usurped #{res[1].join(", ")} routes from #{bound_if}") unless res[1].empty?
       end
     end
+
+    execute "Neutron: add #{bound_if} to #{name}" do
+      command "ovs-vsctl add-port #{name} #{bound_if}"
+      not_if not_if_clause
+    end
     source = ::Nic.new(bound_if)
     # filter out cluster VIPs that a currently assigned to the interface
     addresses = source.addresses.select { |address| my_addresses.include? address }
     routes = source.routes
+
+    not_if_condition = addresses.empty? and routes.empty?
+
+    # Try to bring the cluster back up.
+    if handle_corosync_restart
+      service "start #{node[:corosync][:platform][:service_name]} after network reconfiguration" do
+        service_name node[:corosync][:platform][:service_name]
+        action [:start]
+        not_if { not_if_condition }
+        # ruby_block[wait for cluster to be online] is defined in pacemaker::default
+        notifies :run, "ruby_block[wait for cluster to be online]", :immediately
+      end
+
+      crowbar_pacemaker_sync_mark "sync-neutron-l3_cluster_online" do
+        not_if { not_if_condition }
+      end
+
+      # Clear maintenance flag again on the founder
+      execute "disable maintenance mode" do
+        command "crm --wait configure property maintenance-mode=false"
+        action :run
+        not_if { not_if_condition }
+      end if (CrowbarPacemakerHelper.is_cluster_founder?(node))
+    end
+
     template "/etc/init.d/ovs-usurp-config-#{name}" do
       source "ovs-usurp-config.erb"
       owner "root"
@@ -276,7 +343,7 @@ if neutron[:neutron][:networking_plugin] == 'vmware' or
       # executed for the first time, the physical interface (eth) will not have
       # any addresses or routes assigned anymore. So we should not recreate the
       # init script in that case. Neither should it be removed.
-      not_if { addresses.empty? and routes.empty? }
+      not_if { not_if_condition }
     end
     service "ovs-usurp-config-#{name}" do
       # Don't start it here. It only needs to be executed during boot.
@@ -304,8 +371,6 @@ else
 end
 
 include_recipe "neutron::common_config"
-
-neutron_l3_ha = node.roles.include?("neutron-l3") && node[:neutron][:ha][:l3][:enabled]
 
 service neutron_agent do
   supports :status => true, :restart => true
--- a/chef/cookbooks/neutron/templates/default/ovs-usurp-config.erb
+++ b/chef/cookbooks/neutron/templates/default/ovs-usurp-config.erb
@@ -6,6 +6,8 @@
 # Should-Start:
 # Required-Stop:     $network openvswitch-switch
 # Should-Stop: 
+# X-Start-Before:    crowbar_join openais
+# X-Stop-After:      crowbar_join openais
 # Default-Start:     3 5
 # Default-Stop:      0 1 2 6
 # Short-Description: Usurping IPs and Routes from <%= @source %> to <%= @dest %>


rhafer:
As noted this is still work in progress and pretty rough. I am pretty sure it still has issues. And still have some doubts if we really should go down this route.

Some random notes:

Currently I only tested this on a separate neutron-l3 cluster, so it needs testing in a setup where everything runs in a single cluster.

What happenes if the node is already in maintenance mode before we enable the cluster wide maintance mode? (Will this trigger fencing again? See: https://bugzilla.suse.com/show_bug.cgi?id=917593)

If the first chef-client run fails on a node when the cluster is already stopped there is a chance that the second run will not start it again.

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -202,10 +203,37 @@
       command "ovs-vsctl add-br #{name}; ip link set #{name} up"
       not_if "ovs-vsctl list-br |grep -q #{name}"
     end
-    execute "Neutron: add #{bound_if} to #{name}" do
-      command "ovs-vsctl del-port #{name} #{bound_if} ; ovs-vsctl add-port #{name} #{bound_if}"
-      not_if "ovs-dpctl show system@#{name} | grep -q #{bound_if}"
```
rhafer: This ```not_if``` command (and all the copied versions of that below) is wrong (there is no ```system@br-fixed``` in recent openvswitch releases). Better use: ```ovs-vsctl ifcase-to-br #{bound_if} | grep -q #{name}```

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -202,10 +203,37 @@
...
+    # For details: https://bugzilla.suse.com/show_bug.cgi?id=915176
+    if (node[:crowbar_wall][:network][:nets]["admin"].last == bound_if &&
+        neutron_l3_ha && node[:corosync][:transport] == "udp" )
+
+      # Put the whole cluster into maintenance mode before shutting down
+      # corosync. As this operation affects all cluster nodes only do it on
+      # the founder.
+      if (CrowbarPacemakerHelper.is_cluster_founder?(node))
+        execute "enable maintenance mode" do
+          command "crm --wait configure property maintenance-mode=true"
```
vuntz: Are you sure this puts the whole cluster in maintenance mode, and not just this node?

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -214,10 +242,58 @@
...
+      crowbar_pacemaker_sync_mark "sync-neutron-l3_after_network_reconfig"
+
+      # FIXME: copied from pacemaker::default, should probably move this into a
+      # Helper
+      ruby_block "wait for cluster to be online" do
+        block do
+          require 'timeout'
+          begin
+            Timeout.timeout(60) do
+              cmd = "crm_mon -1 | grep -qi online"
```
vuntz: This only checks for one node to be online, not the whole cluster.

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -202,10 +203,37 @@
...
+    # If:
+    #   * this network is on the same physical interface as the admin network
+    #   * we're running in HA mode
+    #   * corosync is configured to use multicast networking
+    # We need to stop openais/corosync before moving the network interface
+    # into the openvswitch. Otherwise we'll break corosync's multicast
+    # communication and cause fencing.
+    # For details: https://bugzilla.suse.com/show_bug.cgi?id=915176
+    if (node[:crowbar_wall][:network][:nets]["admin"].last == bound_if &&
+        neutron_l3_ha && node[:corosync][:transport] == "udp" )
```
vuntz: I'd use a variable for this boolean value, so that it's super-clear the two `if` are for the same case.

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -214,10 +242,58 @@
         Chef::Log.info("#{name} usurped #{res[1].join(", ")} routes from #{bound_if}") unless res[1].empty?
       end
     end
+
+    execute "Neutron: add #{bound_if} to #{name}" do
+      command "ovs-vsctl del-port #{name} #{bound_if} ; ovs-vsctl add-port #{name} #{bound_if}"
```
vuntz: I'm not sure I get this part: we do a `del-port` and then a `add-port` for what we just deleted?

chef/cookbooks/neutron/recipes/common_agent.rb
rhafer: Yes. For single nodes that's either ```crm node maintenance [node]``` or ```crm configure edit [node] attributes maintenance="true"```

chef/cookbooks/neutron/recipes/common_agent.rb
rhafer: Me neither. AFAIK it's there from the beginning, but I guess it's a good opportunity to cleanup. :)

chef/cookbooks/neutron/recipes/common_agent.rb
rhafer: True. But shouldn't that be enough here? It seems to be enough for the pacemaker::default recipe.

chef/cookbooks/neutron/recipes/common_agent.rb
vuntz: Ah, ok, thanks!

chef/cookbooks/neutron/recipes/common_agent.rb
vuntz: The difference is that in `pacemaker::default`, we just wait for our local pacemaker to be up so that we can start running `crm` commands :-) We don't wait for the whole cluster to be online.

That being said, it's probably good enough here too and maybe I got confused by the fact that we put the whole cluster in maintenance and we just wait for our local pacemaker here.

rhafer: Actually I think it would be better to wait for the whole cluster. Should be possible by adding and addtionally sync-mark here. Provided that it's safe to assume the cluster is back up when every node is able to talk to it's local pacemaker + we have a sync mark after that.

chef/cookbooks/neutron/recipes/common_agent.rb
rhafer: Addressed in: https://github.com/rhafer/barclamp-neutron/commit/ec69587510fab2ee92be03cd741e273d21ae1ebc

chef/cookbooks/neutron/recipes/common_agent.rb
rhafer: Addressed in: https://github.com/rhafer/barclamp-neutron/commit/ec69587510fab2ee92be03cd741e273d21ae1ebc

chef/cookbooks/neutron/recipes/common_agent.rb
rhafer: Addressed in: https://github.com/rhafer/barclamp-neutron/commit/ec69587510fab2ee92be03cd741e273d21ae1ebc

chef/cookbooks/neutron/recipes/common_agent.rb
rhafer: Addressed in: https://github.com/rhafer/barclamp-neutron/commit/ec69587510fab2ee92be03cd741e273d21ae1ebc

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -245,10 +246,41 @@
...
+    #   * we're running in HA mode
+    #   * corosync is configured to use multicast networking
+    # We need to stop openais/corosync before moving the network interface
+    # into the openvswitch. Otherwise we'll break corosync's multicast
+    # communication and cause fencing.
+    # For details: https://bugzilla.suse.com/show_bug.cgi?id=915176
+    handle_corosync_restart = node[:crowbar_wall][:network][:nets]["admin"].last == bound_if &&
+        neutron_l3_ha && node[:corosync][:transport] == "udp"
+
+    not_if_clause = "ovs-vsctl iface-to-br #{bound_if} | grep -q #{name}"
```
vuntz: I'd add an empty line after this one (for readability)

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -245,10 +246,41 @@
...
+          command "crm --wait configure property maintenance-mode=true"
+          not_if not_if_clause
+        end
+      end
+
+      crowbar_pacemaker_sync_mark "sync-neutron-l3_before_network_reconfig" do
+        not_if not_if_clause
+      end
+
+      service "stop openais for network reconfiguration" do
```
vuntz: nitpicking, I guess, but better use `node[:corosync][:platform][:service_name]` than `openais` in this string (same comment later on).

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -257,10 +289,57 @@
...
     end
+
+    execute "Neutron: add #{bound_if} to #{name}" do
+      command "ovs-vsctl add-port #{name} #{bound_if}"
+      not_if not_if_clause
+    end
     source = ::Nic.new(bound_if)
     # filter out cluster VIPs that a currently assigned to the interface
     addresses = source.addresses.select { |address| my_addresses.include? address }
     routes = source.routes
```
vuntz: Would it make sense to move these lines later on? (just for readability)

vuntz: Ah, no, we need `addresses`/`routes` a few lines below.

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -257,10 +289,57 @@
...
+
+    # Try to bring the cluster back up.
+    if handle_corosync_restart
+      service "start openais after network reconfiguration" do
+        service_name node[:corosync][:platform][:service_name]
+        action [:start]
+        not_if { addresses.empty? and routes.empty? }
```
vuntz: Can you clarify this `not_if` bit? Shouldn't we just fail in that case anyway?

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -257,10 +289,57 @@
...
+      end
+
+      crowbar_pacemaker_sync_mark "sync-neutron-l3_after_network_reconfig"
```
vuntz: I'm not sure why we need this sync_mark.

chef/cookbooks/neutron/recipes/common_agent.rb
vuntz: Ah, it's the new `not_if_clause`?

chef/cookbooks/neutron/recipes/common_agent.rb
rhafer: True, shouldn't be needed anymore since I added ```sync-neutron-l3_cluster_online``` further down.

chef/cookbooks/neutron/recipes/common_agent.rb
rhafer: Right. I'll rework that to use ```not_if_clause``` here as well. I use ```addresses.empty? and routes.empty?```  here as an indicator for that we didn't reassign any interface or IP addresses. That should also indicate that corosync wasn't stopped and doesn't need to be restarted. (This should be true for all but the first chef-client run)

But this is also kind of a weak point that I didn't find a better solution for yet. If chef-client fails for what ever reason in between in ```"Have #{name} usurp config from #{bound_if}"``` block and here corosync will never be restarted on a subsequent run. I guess we could make the ```:start``` action unconditional here, it shouldn't do any harm, but I don't really like that either.

vuntz: I'm not sure you should use the same `not_if_clause` because I'm unsure when it's evaluated :-) If it's during the compile phase, then it will work, but if it's in the other phase, then it will break.

And don't worry, corosync will get restarted on subsequent runs since the corosync cookbook should take care of that. Or am I missing something?

rhafer: ```not_if``` is evaluated during the convergence phase AFAIK. But I didn't want reuse the old ```not_if```, I just wanted to assign the new value to it (```addresses.empty? and routes.empty?```) cause it's also used in multiple places.

Regarding the corosync restart, AFAIK the corosync cookbook will think corosync wasn't shutdown cleanly and because of that refuses to start it again, but maybe I am getting this wrong. Additionally there is still the problem that the cluster will be in maintenance mode after that even if it got started …

vuntz: Ah, true about the unclean shutdown. But we can clean this up manually (just remove /var/spool/corosync/block_automatic_start -- iirc, that's the path that is used)

rhafer:
With this state I was able to deploy it successfully a couple of times in different HA setups  (single 2 node cluster, separate 3 node clusters, vlan+gre, ....). -> removing the [WIP] flag

chef/cookbooks/neutron/recipes/common_agent.rb
vuntz: Add `notifies :create, ruby_block["wait for cluster to be online"], :immediately` and get rid of the ruby_block below.

chef/cookbooks/neutron/recipes/common_agent.rb
aspiers: Yep that sounds correct to me too.

aspiers: ... but why do we need to put the whole cluster into maintenance mode?  Why not just this node?

aspiers: (and please put the explanation in comments in the code, not just as a reply in this PR)

rhafer: > ... but why do we need to put the whole cluster into maintenance mode? Why not just this node?

Short answer: https://bugzilla.suse.com/show_bug.cgi?id=917593

I'll try to summarize that in the code.

aspiers: Excellent, thanks :)

chef/cookbooks/neutron/recipes/common_agent.rb
rhafer: Fixed

chef/cookbooks/neutron/recipes/common_agent.rb
rhafer: fixed

chef/cookbooks/neutron/recipes/common_agent.rb
rhafer: Done

chef/cookbooks/neutron/recipes/common_agent.rb
rhafer: Great idea. Implemented. Thanks a lot!

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -257,10 +294,40 @@
...
+    execute "Neutron: add #{bound_if} to #{name}" do
+      command "ovs-vsctl add-port #{name} #{bound_if}"
+      not_if not_if_clause
+    end
     source = ::Nic.new(bound_if)
     # filter out cluster VIPs that a currently assigned to the interface
     addresses = source.addresses.select { |address| my_addresses.include? address }
     routes = source.routes
+
+    not_if_condition = addresses.empty? and routes.empty?
```
vuntz: Tiny detail, but it's slightly confusing to have `not_if_clause` and `not_if_condition`. Can we rename to `not_if_clause_before` and `not_if_clause_after`?

rhafer:
I'll close this for now. We force the unicast mode for corosync currently which doesn't require this complicated cluster restart. We can always reopen it if we need to revisit.

PR #195, opened 2015-02-13
Adjust size of select fields for less empty space

rhafer:


vuntz:
Would it make sense to not hardcode this and instead use something like `length(NeutronService.networking_plugins_valid)`?

rhafer:
> Would it make sense to not hardcode this and instead use something like

Yeah, I tried that. I just gave up on it after failing to succeed  using ```networking_ml2_mechanism_drivers_for_neutron```.

vuntz:
+1

PR #197, opened 2015-02-14
Do not use keystone CLI to fetch service tenant ID

vuntz:
On the initial run, we might not have the keystone client installed
(that's the case if the node has not been used for anything before this
recipe is run).

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_config.rb
+++ b/chef/cookbooks/neutron/recipes/common_config.rb
@@ -89,27 +89,13 @@ nova_notify = {}
 unless nova[:nova].nil? or nova[:nova][:ssl].nil?
   nova_api_host = CrowbarHelper.get_host_for_admin_url(nova, (nova[:nova][:ha][:enabled] rescue false))
   nova_api_protocol = nova[:nova][:ssl][:enabled] ? "https" : "http"
-  keystone_insecure = keystone_settings['insecure'] ? "--insecure" : ""
   nova_insecure = keystone_settings['insecure'] || (nova[:nova][:ssl][:enabled] && nova[:nova][:ssl][:insecure])
-
-  keystone_register "neutron config wakeup keystone" do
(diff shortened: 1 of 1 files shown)


vuntz:
Note that this is really ugly. I was wondering about doing something like

```ruby
package python-keystoneclient
  action :nothing
end.run_action(:install)
```

But I don't like it much either :/

rhafer:
> But I don't like it much either :/

Why not? I think it's a lot better than the approach you implemented. ;)

vuntz:
@rhafer okay, updated for the other approach

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -126,6 +126,11 @@
     action :nothing
   end.run_action(:wakeup)
 
+  # Ensure the keystone call below will not fail because of the missing CLI
+  package python-keystoneclient
```
rhafer: Travis fails. ```do``` is missing here.

vuntz:
Reworked this code path again, after I found out a reboot could break things and lead the node to go into problem state.

chef/cookbooks/neutron/recipes/common_config.rb
vuntz: Oops, sorry, fixed.

dirkmueller:
IMHO thats super ugly.. how about recording the admin tenant id in keystone barclamp and just using the value from chef there? its not going to change anyway, the whole querying that thing dynamically was a mistake to begin with..

vuntz:
@dirkmueller: see https://github.com/crowbar/barclamp-keystone/issues/264 :-)

dirkmueller:
@vuntz can you please rework this after crowbar/barclamp-keystone#264 got merged?

vuntz:
@dirkmueller done, but not tested yet.

rhafer:
+1, survived a quick test

PR #199, opened 2015-02-17
Fix races on 3-node HA clusters

vuntz:
There are two cases where two non-founder nodes can do the same LWRP at the same time, resulting in unexpected failures. I don't think we really need this.

### DIFF
--- a/chef/cookbooks/neutron/recipes/database.rb
+++ b/chef/cookbooks/neutron/recipes/database.rb
@@ -13,6 +13,8 @@
 # limitations under the License.
 #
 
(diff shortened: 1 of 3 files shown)


rhafer:
Even though I still don't quite understand how bnc#918104 actually happend as the pacemaker resources should have already been created by the founder before the other nodes reach that point in the recipes (because of the sync marks). But the changes makes sense anyway so:

+1

vuntz:
@rhafer the issue was not on the initial creation of the resource, but when ensuring that the resource was indeed configured the way it was supposed to be. That's where the non-founders node were doing the same call at the same time.

PR #200, opened 2015-02-17
Enables metadata network

dguitarbite:
This allows the VMs to have custom routes to neutron metadata server.
It will enable the metadata server to accessed from inside the VMs.

### DIFF
--- a/chef/cookbooks/neutron/recipes/l3.rb
+++ b/chef/cookbooks/neutron/recipes/l3.rb
@@ -156,7 +156,7 @@ template "/etc/neutron/dhcp_agent.ini" do
     :dhcp_driver => "neutron.agent.linux.dhcp.Dnsmasq",
     :dhcp_domain => node[:neutron][:dhcp_domain],
     :enable_isolated_metadata => "True",
-    :enable_metadata_network => "False",
+    :enable_metadata_network => "True",
     :nameservers => dns_list
   )
 end


dirkmueller:
@dguitarbite please mention bugreport like " (bnc#123456)" in git commit summary, tia

vuntz:
Is there any drawback to this change, btw?

dguitarbite:
@dirkmueller It works on Cirros for me, for SLES11 images it does not work.

On Cirros image only ping works to server 169.254.169.254. Curl does not work. Were as on SLES images its not working altogehter.

dguitarbite:
@vuntz I cannot seem to recall any major disadvantage by enabling the metadata networks. I will take a look around. There are a few advantages like querying the metadata server eables the VMs to access some information, access to public keypairs etc. as per the configuration. Most common downside to this would be higher load on the network as metadata queries would be consuming some bandwidth (not likely to happen).

vuntz:
@dguitarbite I think there's some misunderstanding. The metadata bits already work in general without this change (in Cirros, for instance). What we're trying to fix is the case where an image already has an expicit route for the network; that's the case in SLES11 images.

dguitarbite:
@vuntz I understand that. I did not get enough time to debug what is the exact issue here.

Metadata is recieved by cloud-init as expected. I am working on explicit manual routes from inside the instace to reach metadata server after the instance is spawned.

dguitarbite:
@vuntz @dirkmueller The metadata server is reachable from within the VM even without this change and I think that this change is not required. Apologizes for this mischap.

PR #201, opened 2015-02-18
HA: Set Optional  ordering for Neutron startup

dirkmueller:
Neutron depends on Keystone and RabbitMQ to be started first


rhafer:
The ```pacemaker_order``` in https://github.com/crowbar/barclamp-neutron/blob/master/chef/cookbooks/neutron/recipes/l3_ha.rb#L147 should also be adapted to use ```crowbar_pacemaker_order_only_existing``` (and include ```rabbitmq```?). 

(Note to self: This depends on https://github.com/crowbar/barclamp-pacemaker/pull/179)

aspiers:
Are we targeting this for GM?  Would require proper QA.

vuntz:
Submitted https://github.com/dirkmueller/barclamp-neutron/pull/1 for my fixes

vuntz:
+1

PR #202, opened 2015-02-24
Use SSL to talk to nova-metadata when SSL is enabled for nova

vuntz:
This depends on https://github.com/crowbar/barclamp-nova/pull/438

### DIFF
--- a/chef/cookbooks/neutron/recipes/l3.rb
+++ b/chef/cookbooks/neutron/recipes/l3.rb
@@ -171,9 +171,6 @@ if novas.length > 0
 else
   nova = node
 end
-# we use an IP address here, and not nova[:fqdn] because nova-metadata doesn't use SSL
-# and because it listens on this specific IP address only (so we don't want to use a name
-# that could resolve to 127.0.0.1).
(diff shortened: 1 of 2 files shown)


chef/cookbooks/neutron/recipes/l3.rb
```diff
@@ -175,7 +175,9 @@
 # and because it listens on this specific IP address only (so we don't want to use a name
 # that could resolve to 127.0.0.1).
 metadata_host = CrowbarHelper.get_host_for_admin_url(nova, (nova[:nova][:ha][:enabled] rescue false))
-metadata_port = "8775"
+metadata_port = nova[:nova][:ports][:metadata] rescue 8775
+metadata_protocol = (nova[:nova][:ssl][:enabled] ? "https" : "http") rescue "http"
```
rhafer: The comment above this code should be updated. Do we also need to switch ```metadata_host``` do you a fqdn when SSL is enabled?

vuntz: Actually, the comment is wrong and we already use the FQDN. I just dropped it.

rhafer:
+1

PR #203, opened 2015-02-27
Make neutron-l3 deployable on SLE12

aplanas:


chef/cookbooks/neutron/recipes/l3.rb
```diff
@@ -53,16 +53,31 @@
...
 
+# Enable ip forwarding on network node for SLE12
+ruby_block "edit /etc/sysctl.d/99-sysctl.conf for net.ipv4.ip_forward" do
+  block do
+    rc = Chef::Util::FileEdit.new("/etc/sysctl.d/99-sysctl.conf")
+    rc.search_file_replace_line(/^net.ipv4.ip_forward =/, 'net.ipv4.ip_forward = 1')
+    rc.write_file
+  end
+  only_if { node[:platform] == "suse" && node[:platform_version].to_f >= 12.0 }
+end
```
vuntz: I guess you need to put this code after `directory "create /etc/sysctl.d for enable-ip_forward" do`, or is the directory just always existing?

aplanas: AFAIK always exists for SLE[SD]12

jsuchome:
+1

rhafer:
@aplanas In genereal this looks good to me. But I'd like to know more about the testing you did on this. Which networking drivers have you tested this with?
Can you probably test this with ml2 + openvswitch + vlan? I am asking because there is some code in the ```common_agent``` recipe that creates a new init script which reassigns IP addresses from physical interfaces to OVS bridges. Can you verify that this really works (and survives a reboot of the l3 node)?

aplanas:
@rhafer Tested with ml2 pluging and linuxbridge (http://crowbar.c16.cloud.suse.de:3000/crowbar/neutron/1.0/proposals/default)

aplanas:
Tested witn ovs + gre

aplanas:
Tested with ovs + vlan in an unpatched cloud instance.

mjura:
:+1: looks good to me

PR #205, opened 2015-03-17
Make awk regexp more specific when fetching id of tenant (bsc#922695)

vuntz:
https://bugzilla.suse.com/show_bug.cgi?id=922695


rhafer:
+1

PR #206, opened 2015-03-18
Add support for VXLAN with ovs

vuntz:


vuntz:
TODO:
- validate linuxbridge support
- make linuxbridge support possible in the UI
- do not hardcode VNI range
- change fixed/floating network creation to use VNI from the user-defined range

vuntz:
Ready for review.

vuntz:
Note that this requires https://github.com/crowbar/barclamp-nova/pull/441

rhafer:
Didn't you want to support VXLAN only with the ovs mech driver? Maybe I am overlooking something, but I don't see that restriction reflected in the patchset currently.  Apart from the commit message :)

rhafer:
Forget about that. The restriction is handled in the UI already. linuxbridge implies VLAN there

vuntz:
@rhafer yeah, the rails side of things forces this. This way, the code is mostly ready for linuxbridge too -- just missing the rails bits and testing.

crowbar_framework/app/assets/javascripts/barclamps/neutron/application.js
```diff
@@ -299,6 +306,7 @@ function ml2_mechanism_drivers_check() {
     $('#ml2_type_drivers_container').hide();
     $('#ml2_type_drivers_default_provider_network_container').hide();
     $('#ml2_type_drivers_default_tenant_network_container').hide();
+    $('#vxlan_group_container').show();
```
rhafer: Isn't this backwards? Shouldn't we hide the vxlan options for now when linuxbridge is selected?

vuntz: This is only for the multicast group setting (which only works for linuxbridge).

In the end, it doesn't matter much for now as, well, we never see the vxlan settings for linuxbridge. But this is to make this ready for when we add linuxbridge support.

chef/cookbooks/neutron/recipes/post_install_conf.rb
```diff
@@ -42,9 +42,10 @@ def mask_to_bits(mask)
 floating_first_ip = IPAddr.new("#{public_range}").to_range().to_a[2]
 floating_last_ip = IPAddr.new("#{public_range}").to_range().to_a[-2]
 floating_pool_start = floating_first_ip if floating_first_ip > floating_pool_start
-
 floating_pool_end = floating_last_ip if floating_last_ip < floating_pool_end
 
+vni_start = [node[:neutron][:vxlan][:vni_start], 0].max
```
aspiers: Just curious what the point of this is?  Why would we need to protect against negative values?

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -149,6 +149,9 @@
 num_vlans = node[:neutron][:num_vlans]
 vlan_end = [vlan_start + num_vlans - 1, 4094].min
 
+vni_start = [node[:neutron][:vxlan][:vni_start], 0].max
+vni_end = [node[:neutron][:vxlan][:vni_end], 16777215].min
```
aspiers: Same here - if the attribute values are out of a valid range, wouldn't it be better to raise an exception, since that should already have been caught by Crowbar validation?

chef/cookbooks/neutron/templates/default/ovs_neutron_plugin.ini.erb
```diff
@@ -64,7 +64,7 @@
 # the server.  In most cases, the default value should be fine.
 #
 # tunnel_bridge = br-tun
-<% if @ml2_type_drivers.include?("gre") -%>
+<% if @ml2_type_drivers.include?("gre") || @ml2_type_drivers.include?("vxlan") -%>
```
aspiers:
```ruby
if (@ml2_type_drivers & %w{gre vxlan}).any?
```

chef/cookbooks/neutron/templates/default/ovs_neutron_plugin.ini.erb
```diff
@@ -125,8 +125,12 @@ bridge_mappings = physnet1:br-fixed
 # Example: tunnel_types = gre
 # Example: tunnel_types = vxlan
 # Example: tunnel_types = vxlan, gre
-<% if @ml2_type_drivers.include?("gre") -%>
+<% if @ml2_type_drivers.include?("gre") && @ml2_type_drivers.include?("vxlan") -%>
```
aspiers: I feel like this should be doable using `case ... when ... end` but maybe it isn't :-/

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -157,6 +159,34 @@ def validate_proposal_after_save proposal
       validation_error("The default tenant network type driver \"#{ml2_type_drivers_default_tenant_network}\" is not a selected ml2 type driver")
     end
 
+    if ml2_type_drivers.include? 'vxlan'
```
aspiers: This method is getting long; it would be nice to split it out into a separate method if possible.

aspiers:
LGTM apart from the nitpicks, but then I'm no networking guru and only just learned the details of VXLAN today ...

nkrinner:
+1

chef/cookbooks/neutron/recipes/post_install_conf.rb
vuntz: @aspiers it's our old debate we have all the time ;-) I'm generally writing code that tries to be safe with invalid input.

chef/cookbooks/neutron/recipes/server.rb
vuntz: (see above; the point is that the attribute can be changed outside of crowbar, directly in chef; just the way I write code again)

chef/cookbooks/neutron/templates/default/ovs_neutron_plugin.ini.erb
vuntz: I don't think the first test fits a `case`, unfortunately. At least, I don't see how to make it fit :-)

chef/cookbooks/neutron/templates/default/ovs_neutron_plugin.ini.erb
vuntz: Any other opinion on this? I find the current code more readable, but I'm happy to change if others agree with @aspiers that's it's cleaner the other way.

crowbar_framework/app/models/neutron_service.rb
vuntz: You're right, done!

chef/cookbooks/neutron/recipes/server.rb
aspiers: Hmm, I don't remember our previous discussions but I can't really agree in this case.  If I changed a value via Chef and accidentally (or through misunderstanding of the attribute) provided an invalid value, I would much prefer to be told about it then have that value silently ignored, because the latter behaviour is more likely to result in confusion from unexpected side-effects later on.  And noone except developers should be changing values via Chef anyway.  But if that doesn't persuade you then I guess let's just agree to disagree and go with your version; it's not a big deal.

chef/cookbooks/neutron/templates/default/ovs_neutron_plugin.ini.erb
aspiers: Me neither, at least not without creating a new class and overloading `#===`.

chef/cookbooks/neutron/templates/default/ovs_neutron_plugin.ini.erb
aspiers: There's not much to choose between the two versions: use of `&` and `any?` is ever so slightly cryptic, but OTOH repetition of `@ml2_type_drivers.include?` is ugly and requires extra effort for the brain and eyes to check that it really is exact repetition.  What swings it for me is the scalability of my version: if other cases need to be added later, then the temptation to copy'n'paste resulting in a 3rd `@ml2_type_drivers.include?()` call will be too great to resist :)

vuntz: I don't find the current way ugly, to be honest and it requires less effort to my brain ;-) But really, I don't care in the end and can live with both ways.

I just want to be sure others are fine with the slightly more cryptic way too.

aspiers: IIRC from a conversation a long time ago, that's the same crazy brain which thinks that

    "... \"foo bar\" \"baz qux\" ..."

    is more readable than

    '... "foo bar" "baz qux" ...'

;-p  Just goes to show that personal preferences can be very different.  But I also don't mind either way (as long as no more copy'n'paste happens in the future) so I'll leave it to you to choose ;-)

tboerger: > if (@ml2_type_drivers & %w{gre vxlan}).any?

I dislike this, the current code is better readable especially as many people don't know what the ```&``` operator does in this case.

rhafer: > > if (@ml2_type_drivers & %w{gre vxlan}).any?
> 
> I dislike this, the current code is better readable especially as many people don't know what the & operator does in this case.

Yeah, I agree with that.

chef/cookbooks/neutron/templates/default/ovs_neutron_plugin.ini.erb
tboerger: Maybe it's better readable to select ```gre``` and ```vxlan``` from the array and simply join it?

tboerger:
LGTM

chef/cookbooks/neutron/attributes/default.rb
```diff
@@ -31,6 +31,10 @@
 default[:neutron][:network][:gre_start] = 1
 default[:neutron][:network][:gre_stop] = 1000
 
+default[:neutron][:vxlan][:vni_start] = 4096
+default[:neutron][:vxlan][:vni_stop] = 99999
+default[:neutron][:vxlan][:multicast_group] = "239.1.1.1"
+
```
toabctl: Why don't you use the same naming schema as for gre? like ```default[:neutron][:network][:vxlan_vni_start]```. I think your version is nicer but inconsistent with the other code. Isn't it?

vuntz: That's because the GRE attributes are actually not part of the schema. I could rename them in a later PR, though, if you want (and add them to the schema).

chef/cookbooks/neutron/templates/default/ovs_neutron_plugin.ini.erb
vuntz: Why not. This would look like this:
```ruby
@ml2_type_drivers.select { |t| ["vxlan", "gre"].include?(t) }.join(",")
```

Opinions?

(actually, it'd be slightly different as I'd make this a template variable, so we don't write an empty tunnel_types)

tboerger: A template variable sounds good to me.

chef/cookbooks/neutron/templates/default/ovs_neutron_plugin.ini.erb
aspiers: People should learn the Ruby operators then :)

chef/cookbooks/neutron/templates/default/ovs_neutron_plugin.ini.erb
aspiers: Oh yeah, that's much nicer.

vuntz: Done.

chef/cookbooks/neutron/attributes/default.rb
vuntz: I added a commit for this.

vuntz:
There were some +1, but since the code changed a bit based on feedback, could it be possible to get some fresh +1 after a new review?

aspiers:
LGTM at a very quick glance

chef/cookbooks/neutron/templates/default/ovs_neutron_plugin.ini.erb
tboerger: Even me as a ruby developer for multiple years never used this :)

aspiers: Pretty sure I remember reading about `Array#&` in about 2001.  That rich out-of-the-box API is one of the things which most attracted me to Ruby in the early days.  But I guess it's much less likely you would need `Array` intersection if you mainly do web development.

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -104,6 +106,50 @@ def create_proposal
...
+  def validate_gre gre_settings
+    if gre_settings["tunnel_id_start"] < 1 || gre_settings["tunnel_id_start"] > 2147483647
+      validation_error("Start of GRE tunnel ID range must be between 0 and 2147483647")
+    end
+    if gre_settings["tunnel_id_end"]  < 1 || gre_settings["tunnel_id_end"] > 2147483647
+      validation_error("End of GRE tunnel ID range must be between 0 and 2147483647")
+    end
+    if gre_settings["tunnel_id_start"] > gre_settings["tunnel_id_end"]
+      validation_error("End of GRE tunnel ID range must be higher than start of GRE tunnel ID range")
+    elsif gre_settings["tunnel_id_start"] + 1 > gre_settings["tunnel_id_end"]
```
rhafer: Doesn't this mean ```gre_settings["tunnel_id_start"] == gre_settings["tunnel_id_end"]```, at least if you can the previous ```if  gre_settings["tunnel_id_start"] > gre_settings["tunnel_id_end"]``` into account?

crowbar_framework/app/views/barclamp/neutron/_edit_attributes.html.haml
```diff
@@ -40,6 +40,17 @@
...
+        %fieldset
+          %legend
+            = t(".vxlan_header")
+          = integer_field %w(vxlan vni_start)
+          = integer_field %w(vxlan vni_end)
+          #vxlan_group_container
+            = string_field %w(vxlan multicast_group)
+            %span.help-block
+              = t('.vxlan.multicast_group_hint')
+
```
rhafer: Since you also added the gre attributes, should we also offer them in the UI?

crowbar_framework/app/models/neutron_service.rb
vuntz: Hrm, no, it means that we do not accept `gre_settings["tunnel_id_start"] == gre_settings["tunnel_id_end"]`. The reason I did two tests is really for the different error messages.

rhafer: Yes, I know. But I think using ```gre_settings["tunnel_id_start"] == gre_settings["tunnel_id_end"]``` in the ```elsif``` is somewhat more readable than ```gre_settings["tunnel_id_start"] + 1 > gre_settings["tunnel_id_end"]```

crowbar_framework/app/views/barclamp/neutron/_edit_attributes.html.haml
vuntz: @rhafer: done

crowbar_framework/app/models/neutron_service.rb
vuntz: Ah, sorry, I misunderstood your comment. Done (also for vxlan check)

rhafer:
I think I am fine with this now. +1

aplanas:
https://github.com/crowbar/barclamp-neutron/pull/209/ and https://github.com/crowbar/barclamp-neutron/pull/210/ share some older(?) version of https://github.com/vuntz/barclamp-neutron/commit/d879f288429ecba5a4def3e41732635650dacedd and https://github.com/vuntz/barclamp-neutron/commit/378ae48cf8be42e3f41e1c816a5f97ff575aef15, while https://github.com/vuntz/barclamp-neutron/commit/09d6e95e66c91e52ebbafe653ca5f965dce2ae2e and https://github.com/vuntz/barclamp-neutron/commit/bf7160378f2a037329358f85643f662fb403b9cc are missing.

But +1 for this.

vuntz:
@aplanas I'll rebase the other PR once this one is merged. With the latest +1, I guess I'll just do it now actually. Thanks everyone!

PR #207, opened 2015-03-19
use lower MTU value for GRE tunnels (bnc#917328)

bmwiedemann:
this avoids fragmentation as per
http://docs.openstack.org/admin-guide-cloud/content/openvswitch_plugin.html

and avoids networking problems with SLE12 compute hosts and GRE
https://bugzilla.suse.com/show_bug.cgi?id=917328

### DIFF
--- a/chef/cookbooks/neutron/templates/default/dnsmasq-neutron.conf.erb
+++ b/chef/cookbooks/neutron/templates/default/dnsmasq-neutron.conf.erb
@@ -1,0 +1,4 @@
+# this file is managed by chef to give extra options to dnsmasq
+<% if @force_mtu -%>
+dhcp-option-force=26,<%= @force_mtu.to_s %>
+<% end -%>
--- a/chef/cookbooks/neutron/recipes/network_agents.rb
+++ b/chef/cookbooks/neutron/recipes/network_agents.rb
@@ -125,6 +125,18 @@ template "/etc/neutron/metering_agent.ini" do
   )
 end
 
+ml2_drivers = node[:neutron][:ml2_type_drivers]
+template "/etc/neutron/dnsmasq-neutron.conf" do
+  source "dnsmasq-neutron.conf.erb"
+  owner "root"
+  group "root"
+  mode "0644"
+  variables(
+    # nil means to leave the default MTU
+    :force_mtu => ml2_drivers.include?('gre') || ml2_drivers.include?('vxlan') ? 1400 : nil
+  )
+end
+
 dns_list = node[:dns][:forwarders].join(",")
 
 template "/etc/neutron/dhcp_agent.ini" do
@@ -189,6 +201,7 @@ service node[:neutron][:platform][:dhcp_agent_name] do
   action [:enable, :start]
   subscribes :restart, resources("template[/etc/neutron/neutron.conf]")
   subscribes :restart, resources("template[/etc/neutron/dhcp_agent.ini]")
+  subscribes :restart, resources("template[/etc/neutron/dnsmasq-neutron.conf]")
   provider Chef::Provider::CrowbarPacemakerService if ha_enabled
 end
 
--- a/chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
+++ b/chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
@@ -70,7 +70,7 @@ enable_metadata_network = <%= @enable_metadata_network %>
 dhcp_domain = <%= @dhcp_domain %>
 
 # Override the default dnsmasq settings with this file
-# dnsmasq_config_file =
+dnsmasq_config_file = /etc/neutron/dnsmasq-neutron.conf
 
 # Comma-separated list of DNS servers which will be used by dnsmasq
 # as forwarders.


chef/cookbooks/neutron/recipes/l3.rb
```diff
@@ -156,6 +156,16 @@
   )
 end
 
+template "/etc/neutron/dnsmasq-neutron.conf" do
+  source "dnsmasq-neutron.conf.erb"
+  owner "root"
+  group "root"
+  mode "0644"
+  variables(
+    :gre => true
```
bmwiedemann: this needs some improvement by a chef/neutron expert

vuntz:
```
:low_mtu => node[:neutron][:ml2_type_drivers].include? 'gre'
```

chef/cookbooks/neutron/templates/default/dnsmasq-neutron.conf.erb
```diff
@@ -0,0 +1,3 @@
+<% if @gre -%>
+dhcp-option-force=26,1458
+<% end -%>
```
vuntz: Maybe add a comment so that when gre is not used, this file is not totally empty and confusing?

chef/cookbooks/neutron/recipes/l3.rb
```diff
@@ -156,6 +156,16 @@
   )
 end
 
+template "/etc/neutron/dnsmasq-neutron.conf" do
+  source "dnsmasq-neutron.conf.erb"
+  owner "root"
+  group "root"
+  mode "0644"
+  variables(
+    :low_mtu => node[:neutron][:ml2_type_drivers].include?('gre')
```
bmwiedemann: is the :ml2_type_drivers value always available?

rhafer: It's an attribute from the proposal (marked as "required" in the schema). So yes, assuming that it's allways there should be fine. (It might be empty though, but doesn't matter here).

chef/cookbooks/neutron/templates/default/dnsmasq-neutron.conf.erb
```diff
@@ -0,0 +1,4 @@
+# this file is managed by chef to give extra options to dnsmasq
+<% if @low_mtu -%>
+dhcp-option-force=26,1458
```
vuntz: Upstream recommends 1400, not 1458. Do you know why?

bmwiedemann: I guess they found that it worked and did not want to spend time trying many different values to find the largest working one.

bmwiedemann:
I successfully tested that update works as is using
cd /opt/dell/
curl https://github.com/crowbar/barclamp-neutron/pull/207.patch | patch -p1
knife cookbook upload -o /opt/dell/chef/cookbooks/ neutron
crowbar neutron proposal commit default # to trigger chef-client runs

chef/cookbooks/neutron/templates/default/dnsmasq-neutron.conf.erb
vuntz: So I guess it'd be useful to know how you reached 1458 to decide. I'm not sure where the 42 overhead comes from (except for the experimental approach where you try it). GRE can be 4, 8 or 12 bytes, IPv4 is 20 bytes, IPv6 (assuming we switch to IPv6 later) would be 40 bytes instead of 20 bytes.

bmwiedemann: I found the 1458 by testing with ssh and netcat transfers. size calculation might be IP+Eth+VLAN+GRE = 20 + 14 + 4 + 4
we can give it another 40 bytes to be on the safe (but slower) side

rossella:
LGTM!

rhafer:
I think we should really go with the recommended 1400. AFAICS:

GRE Headers can be up to 16 bytes
Ethernet (excluding VLAN) is 6 SMAC+6 DMAC +2 TYPE + 4 CRC = 18
VLAN is 4 byte
and 
IPv4 headers can be up to 60 bytes (IPv6 only 40)
that makes: 16 + 18 + 4 + 60 = 98 which is pretty close to the 1400 that upstream recommends.

chef/cookbooks/neutron/recipes/l3.rb
rhafer: We recently merged support for VXLAN I guess this also requires MTU adjustments.  AFAICS the VXLAN Header has a constant size of 8 bytes.

bmwiedemann:
I just checked sizes in an actual cloud-tcpdump: GRE=8, Ethernet=14, VLAN=4, IP=20
making a total of 46 so we still have plenty room for tcp-options (which seem to be rarely used)

rhafer:
If we really want to fix/workaround the above bug we need to use save settings. So can we please just use 1400 like recommended upstream? (I can see no prove that that number came up just because of guess work)

If you think we need to optimize this for the last byte I suggest we offer an attribute for the guest MTU in the proposal and default that to 1400 when using GRE or VXLAN.
We might need to add such an attribute anyway at some point if we ever add support for Jumbo Frames.

bmwiedemann:
I updated the PR to make MTU variable and use 1400 for now, even though I consider it bad cargo cult.

I tried to find the source of the 1400 using git blame and traced it through 4 renames to
https://github.com/openstack/openstack-manuals/commit/1fe3fe0a3199017aaabf3a
which does not tell how the 1400 was calculated.

https://ask.openstack.org/en/question/6140/quantum-neutron-gre-slow-performance/ predates it... and has mentions that 1454 works too (likely because of the 46 byte overhead from GRE).

chef/cookbooks/neutron/recipes/l3.rb
```diff
@@ -156,6 +156,16 @@
   )
 end
 
+template "/etc/neutron/dnsmasq-neutron.conf" do
+  source "dnsmasq-neutron.conf.erb"
+  owner "root"
+  group "root"
+  mode "0644"
+  variables(
+    :mtu => node[:neutron][:ml2_type_drivers].include?('gre') ? 1400 : 1500
```
rhafer: You're still missing the "vxlan" case here. We need the lower MTU in that case as well.

bmwiedemann:
updated to handle VXLAN too

chef/cookbooks/neutron/templates/default/dnsmasq-neutron.conf.erb
```diff
@@ -0,0 +1,3 @@
+# this file is managed by chef to give extra options to dnsmasq
+dhcp-option-force=26,<%= @mtu.to_s %>
+
```
vuntz: (no need for the empty line)

chef/cookbooks/neutron/templates/default/dnsmasq-neutron.conf.erb
```diff
@@ -0,0 +1,3 @@
+# this file is managed by chef to give extra options to dnsmasq
+dhcp-option-force=26,<%= @mtu.to_s %>
```
vuntz: I thought we wanted to not set any dhcp option if there's no need for that?

bmwiedemann: I thought, it might be used later for other cases, e.g. let user give a mtu

rhafer: > I thought, it might be used later for other cases, e.g. let user give a mtu
Even in that case we probably only want to set that option when using a non-default value.
So IMO we should add an ```only_if``` condition to the ```template /etc/neutron/dnsmasq-neutron.conf``` and add a flag to the ```dhcp_agent.ini``` to only include that file when needed.

bmwiedemann:
updated to not use a force value in the non-GRE/VXLAN case

rhafer:
+1

vuntz:
This needs to be rebased :/

bmwiedemann:
rebased it. git nicely auto-resolved the commit with a 3-way-merge.

PR #208, opened 2015-03-20
Fix races on 3-node HA clusters

vuntz:
Backported from tex.


rhafer:
+1 (and merging since this is a backport that got approved already for master)

PR #209, opened 2015-03-23
Enable L2 population mechanism driver

vuntz:
This is only possible when VXLAN or GRE is used.

For more details about L2 population, see
http://docs.openstack.org/admin-guide-cloud/content/ml2_l2pop_scenarios.html
and https://wiki.openstack.org/wiki/L2population

This comes on top of https://github.com/crowbar/barclamp-neutron/pull/206

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_agent.rb
+++ b/chef/cookbooks/neutron/recipes/common_agent.rb
@@ -108,6 +108,8 @@ if neutron[:neutron][:networking_plugin] == "ml2"
     end
   end
 
+  ml2_type_drivers = neutron[:neutron][:ml2_type_drivers]
+
   case
   when ml2_mech_drivers.include?("openvswitch")
     directory "/etc/neutron/plugins/openvswitch/" do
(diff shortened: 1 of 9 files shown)


vuntz:
There's the question of whether this should be something that can be disabled. I'm not sure why people would want that, though.

@rossella: any opinion on whether it can be useful to disable L2 population?

rossella:
@vuntz L2 population basically populates the flows to avoid ARP requests to be broadcasted so I don't see any advantage in disabling it. In addition to that l2 population is required for DVR, so let's enable it :)

vuntz:
Trying this and, hrm, this seems to kill networking :-) Not sure why yet.

vuntz:
Ok, this seems to work now. It seems I had a weird mix of old patch/new patch. I'm not totally sure that this won't break anything on existing setups, so will reinstall from scratch to test that...

vuntz:
Works now.

One bit that I'm wondering, though, is how to handle that from an update perspective: due to the fact that the changes in the config files will only happen with the periodic chef-client run here, not all config files will be updated at the same time and some agents might stop working until neutron-server is restarted.

So in the worst case, that could be a 15 minutes downtime for networking.

Hence I'm mixed: should we try to be nice and not force l2population on users (which means we need an attribute for this, that would be set to false for old users on migration, and to true for new users)? Or do we believe it's acceptable?

jsuchome:
Codewise, looks good, correctly implements what is requested by documentation. So that would be + 1 from me (but not tested)

aplanas:
@vuntz I would go for the first option with documentation about how to manually force a fast update (via knife) to the different agents.

The second option will be of no use in Cloud 6, and I feel that differentiate the upgrade case from re-applying the barclamp and other scenarios can be tricky to done it right. But this is an uninformed guess.

aplanas:
+1

rhafer:
> Hence I'm mixed: should we try to be nice and not force l2population on users (which means we need an attribute for this, that would be set to false for old users on migration, and to true for new users)? Or do we believe it's acceptable?

As we're going to release this as an maint. update I think forcing this on the users and causing network downtime is not acceptable. So I'd vote for an additional attribute. But I'd suggest to defaulting to "true" in the template (to have it enabled for new deployments) while setting it to "false" in the existing proposals/roles.

For the next big release we should probably update all installations to use l2pop and remove that attribute again.

vuntz:
@rhafer added a commit for that.

rhafer:
+1

aplanas:
+1

PR #210, opened 2015-03-24
Add support for DVR

vuntz:
**Please note that this PR includes https://github.com/crowbar/barclamp-neutron/pull/209**

*First commit:*
As the L3 agent will also run on compute host, this name makes no sense.
And we really reference such nodes as network nodes, so it makes more
sense this way.

*Second commit:*
DVR stands for Distributed Virtual Routers, see
https://github.com/openstack/neutron-specs/blob/master/specs/juno/neutron-ovs-dvr.rst
https://wiki.openstack.org/wiki/Neutron/DVR

The implementation is loosely based on what's in
https://wiki.openstack.org/wiki/Neutron/DVR/HowTo

For DVR, the L3 agent needs to run not just on the network node, but
also on compute nodes. So the configuration for this has to move to the
common code path.

When using HA, we still setup the L3 agent with pacemaker on the network
nodes, but it's unclear yet if this just works fine.

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_agent.rb
+++ b/chef/cookbooks/neutron/recipes/common_agent.rb
@@ -312,12 +312,12 @@ end
 
 include_recipe "neutron::common_config"
 
-neutron_l3_ha = node.roles.include?("neutron-l3") && node[:neutron][:ha][:l3][:enabled]
+neutron_network_ha = node.roles.include?("neutron-network") && node[:neutron][:ha][:network][:enabled]
 
 service neutron_agent do
   supports :status => true, :restart => true
   action [:enable, :start]
   subscribes :restart, resources("template[#{agent_config_path}]")
   subscribes :restart, resources("template[/etc/neutron/neutron.conf]")
-  provider Chef::Provider::CrowbarPacemakerService if neutron_l3_ha
+  provider Chef::Provider::CrowbarPacemakerService if neutron_network_ha
 end
--- a/chef/cookbooks/neutron/recipes/common_agent.rb
+++ b/chef/cookbooks/neutron/recipes/common_agent.rb
@@ -130,7 +130,8 @@ if neutron[:neutron][:networking_plugin] == "ml2"
       variables(
         :ml2_type_drivers => ml2_type_drivers,
         :tunnel_types => ml2_type_drivers.select { |t| ["vxlan", "gre"].include?(t) },
-        :use_l2pop => neutron[:neutron][:use_l2pop] && (ml2_type_drivers.include?("gre") || ml2_type_drivers.include?("vxlan"))
+        :use_l2pop => neutron[:neutron][:use_l2pop] && (ml2_type_drivers.include?("gre") || ml2_type_drivers.include?("vxlan")),
+        :dvr_enabled => neutron[:neutron][:use_dvr]
       )
     end
   when ml2_mech_drivers.include?("linuxbridge")
@@ -312,7 +313,7 @@ end
 
 include_recipe "neutron::common_config"
 
-neutron_network_ha = node.roles.include?("neutron-network") && node[:neutron][:ha][:network][:enabled]
+neutron_network_ha = node.roles.include?("neutron-network") && neutron[:neutron][:ha][:network][:enabled]
 
 service neutron_agent do
   supports :status => true, :restart => true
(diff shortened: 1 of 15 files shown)


vuntz:
The role renaming is something we have to be careful on validation.

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -305,12 +309,65 @@
...
   subscribes :restart, resources("template[#{agent_config_path}]")
   subscribes :restart, resources("template[/etc/neutron/neutron.conf]")
-  provider Chef::Provider::CrowbarPacemakerService if neutron_l3_ha
+  provider Chef::Provider::CrowbarPacemakerService if neutron_network_ha
+end
+
+# Note: if vmware, we do not reach this code
+
+if neutron[:neutron][:dvr_enabled] || node.roles.include?("neutron-network")
+  unless node[:neutron][:use_gitrepo]
```
nkrinner: I think neutron[:neutron][:dvr_enabled] is never defined or set.

vuntz: Ah, good catch, fixed.

vuntz:
Note that there's no migration of old routers in Juno yet, see https://bugs.launchpad.net/neutron/+bug/1348309

So for testing, we have to test with fresh new routers :-)

vuntz:
So with this patch, and with DVR **disabled**, things still work. I'll check with DVR now.

vuntz:
Note that we need https://github.com/crowbar/barclamp-nova/pull/443 for DVR to work (thanks @rossella for the info!)

vuntz:
Due to the fact that the metadata agent setup had to move to the common_agent recipe, there was quite some code reorg in the latest force-push. Sorry for that :/

vuntz:
For the record, it seems to be working fine, except that floating IP don't work (yet).

vuntz:
Looks like this works fine (with https://build.opensuse.org/package/view_file/Cloud:OpenStack:Juno:Staging/openstack-neutron/0001-Use-j-NOTRACK-instead-of-j-CT-notrack.patch?expand=1)

vuntz:
I'll need to rebase on top of the l2pop PR that changed.

vuntz:
Rebased on top of the l2pop PR that changed.

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -306,14 +231,204 @@
...
+    template agent_config_path do
+      cookbook "neutron"
+      source "ovs_neutron_plugin.ini.erb"
+      owner "root"
+      group node[:neutron][:platform][:group]
+      mode "0640"
+      variables(
+        :ml2_type_drivers => ml2_type_drivers,
+        :tunnel_types => ml2_type_drivers.select { |t| ["vxlan", "gre"].include?(t) },
+        :use_l2pop => neutron[:neutron][:use_l2pop] && (ml2_type_drivers.include?("gre") || ml2_type_drivers.include?("vxlan"))
```
aplanas: Missing a comma here (at the end)

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -306,14 +231,204 @@
...
+    neutron_agent_pkg = node[:neutron][:platform][:lb_agent_pkg]
+    agent_config_path = "/etc/neutron/plugins/linuxbridge/linuxbridge_conf.ini"
+    interface_driver = "neutron.agent.linux.interface.BridgeInterfaceDriver"
+    external_network_bridge = ""
+  end
 
-service neutron_agent do
+  # L2 agent
+  unless neutron[:neutron][:use_gitrepo]
+    package neutron_agent_pkg
```
aplanas: I think that you are moving this to the low part of the file, but there is a service that I think is related with this package, that is addressed before: `service_name "openstack-neutron-ovs-cleanup" if %w(suse).include?(node.platform)`

vuntz: Hrm, you're right.

I guess I can't avoid some ugly mess in the code...

vuntz: Fixed

chef/cookbooks/neutron/recipes/common_agent.rb
vuntz: Gah, I broke that during the rebase. Thanks, fixed!

aplanas:
+1 Tested in c16

jsuchome:
+1 tested at c13 (together with nova PR), seems to work well

aplanas:
Makes sense to test it in a HyperV installation, to make sure that the DVR do not break in this case?

vuntz:
> Makes sense to test it in a HyperV installation, to make sure that the DVR do not break in this case?

As HyperV requires vlan only, and this works with vxlan, I'd say no :-)

vuntz:
I force pushed to fix https://github.com/vuntz/barclamp-neutron/commit/4bf0ad38f2e41e10f6475cdf1e616ec85fa3434d#commitcomment-10513123 but there's no change in the code at the end of the commits.

rossella:
For some reason I can't comment on the line.
Anyway DVR does support GRE too, see [1] (didn't find any better link) "In Juno the VXLAN/GRE network type underlays were enabled to be supported as DVR Interfaces."

[1] https://blueprints.launchpad.net/neutron/+spec/neutron-ovs-dvr-vlan

aplanas:
@rossella the funny thing is that looks like that works in VLAN. You can see it in `c16`, with VXLAN, GRE and two VLAN networks connected at the same distributed router.

rossella:
@aplanas with VLAN is supposed to work in Kilo, not before...

aplanas:
@rossella indeed! This is why `c16` is suspicious (also @vuntz tested it first in a different setup).

vuntz:
@rossella I added a commit to amend the validation and accept GRE when there's no VXLAN.

PR #212, opened 2015-04-03
webui: Automatically enable/disable l2pop depending on used types

vuntz:
L2 population doesn't work with vlan-only, so we need to disable it in
that case. But we still want to use it by default when possible, so use
some javascript magic.


rhafer:
+1

rossella:
+1

PR #214, opened 2015-04-22
allow nova-access to work with ssl-keystone (bnc#895594)

bmwiedemann:
This is needed when keystone is deployed with SSL certs
that dont cover all internal hostnames (aka insecure),
but nova is not deployed with HTTPS,
so the insecure flag is missing there.

https://bugzilla.suse.com/show_bug.cgi?id=895594

This is very much like https://github.com/crowbar/barclamp-nova/pull/445


bmwiedemann:
This was making nova boot work in a quick test

vuntz:
+1

rhafer:
+1

PR #215, opened 2015-05-04
[post-tex] Drop PFS support (and some cleanup)

vuntz:
Pull from source hasn't been maintained for quite some time and it's
very likely broken. So drop support for it.

### DIFF
--- a/chef/cookbooks/neutron/recipes/network_agents.rb
+++ b/chef/cookbooks/neutron/recipes/network_agents.rb
@@ -16,33 +16,12 @@
 
 include_recipe "neutron::common_agent"
 
-unless node[:neutron][:use_gitrepo]
-  pkgs = [ node[:neutron][:platform][:dhcp_agent_pkg], node[:neutron][:platform][:metering_agent_pkg] ]
-  pkgs.uniq.each { |p| package p }
+pkgs = [ node[:neutron][:platform][:dhcp_agent_pkg], node[:neutron][:platform][:metering_agent_pkg] ]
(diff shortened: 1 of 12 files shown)


chef/cookbooks/neutron/recipes/network_agents.rb
```diff
@@ -16,34 +16,13 @@
 
 include_recipe "neutron::common_agent"
 
-unless node[:neutron][:use_gitrepo]
-  pkgs = [ node[:neutron][:platform][:dhcp_agent_pkg], node[:neutron][:platform][:metering_agent_pkg] ]
-  pkgs.uniq.each { |p| package p }
+pkgs = [ node[:neutron][:platform][:dhcp_agent_pkg], node[:neutron][:platform][:metering_agent_pkg] ]
+pkgs.uniq.each { |p| package p }
```
jsuchome: Wouldn't it be cleaner/nicer, and more in line with next package call, just to write
``package node[:neutron][:platform][:dhcp_agent_pkg]``
``package node[:neutron][:platform][:metering_agent_pkg]``

(missing ``uniq`` should not hurt)

jsuchome:
+1 (tested few neutron commands after applying the patch and redeploying barclamp)

chef/cookbooks/neutron/recipes/network_agents.rb
vuntz: Fixed with a new commit.

ovisan:
:+1:

PR #217, opened 2015-05-06
pass keystone admin password to neutron-ha-tool via file (bsc#922751)

aspiers:
Will need backporting to tex and maybe stoney too.

Requires corresponding change in neutron-ha-tool:

https://build.suse.de/request/show/57156

### DIFF
--- a/chef/cookbooks/neutron/recipes/network_agents_ha.rb
+++ b/chef/cookbooks/neutron/recipes/network_agents_ha.rb
@@ -139,6 +139,7 @@ pacemaker_primitive ha_tool_primitive_name do
   agent node[:neutron][:ha][:network][:ha_tool_ra]
   params ({
     "os_auth_url"    => keystone_settings["internal_auth_url"],
+    "os_region_name" => keystone_settings["endpoint_region"],
     "os_tenant_name" => keystone_settings["admin_tenant"],
     "os_username"    => keystone_settings["admin_user"],
     "os_insecure"    => keystone_settings["insecure"] || node[:neutron][:ssl][:insecure]
--- a/chef/cookbooks/neutron/recipes/network_agents_ha.rb
+++ b/chef/cookbooks/neutron/recipes/network_agents_ha.rb
@@ -125,13 +125,22 @@ ha_tool_primitive_name = "neutron-ha-tool"
 # right CA file as we allow Keystone's and Neutron's to use different CAs.  So
 # we just rely on the correct CA files being installed in a system wide default
 # location.
+file "/etc/neutron/os_password" do
+  owner 'root'
+  group 'root'
+  mode '0600'
+  content keystone_settings["admin_password"]
+  # Our Chef is apparently too old for this :-/
+  #sensitive true
+  action :create
+end
+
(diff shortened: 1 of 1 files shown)


aspiers:
Tested and it works, but haven't tested the resulting file with neutron-ha-tool yet.

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -125,13 +125,22 @@
...
+  action :create
+end
+
 pacemaker_primitive ha_tool_primitive_name do
   agent node[:neutron][:ha][:network][:ha_tool_ra]
   params ({
     "os_auth_url"    => keystone_settings["internal_auth_url"],
     "os_tenant_name" => keystone_settings["admin_tenant"],
     "os_username"    => keystone_settings["admin_user"],
-    "os_password"    => keystone_settings["admin_password"],
```
vuntz: Looks good to me, but shouldn't there be a way in the OCF to specify the path to the file?

vuntz: Reason I'm asking is that I'm unsure I'd put it in /etc/neutron/os_password -- I'd probably make it clearer that it's a file for the ha tool...

aspiers: Well actually in the long term it's not necessarily just for the HA tool; it could be for anything which needs to communicate with the API - possibly even stuff outside neutron, e.g. the keystone OCF RA.  So you *could* argue it should be in `/etc/keystone`.  But right now it really shouldn't matter at all.

I can't imagine any reason why it would need to be a configurable parameter, so I suggest we stick with the simplest thing which could possibly work for now.  If we need to make it configurable later then we can easily do that, but that seems like a waste of effort right now.

jsuchome:
This works as planned +1

aspiers:
Added a patch for region name.

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -125,13 +125,23 @@
...
+end
+
 pacemaker_primitive ha_tool_primitive_name do
   agent node[:neutron][:ha][:network][:ha_tool_ra]
   params ({
     "os_auth_url"    => keystone_settings["internal_auth_url"],
+    "os_region_name" => keystone_settings["endpoint_region"],
     "os_tenant_name" => keystone_settings["admin_tenant"],
     "os_username"    => keystone_settings["admin_user"],
-    "os_password"    => keystone_settings["admin_password"],
```
nkrinner: This breaks the code for me and I don't understand why this change is needed.

aspiers: It's explained in the commit message. Storing the password on disk is the whole point of this PR.

aspiers: It probably breaks because you don't have the patched neutron-ha-tool.  Strongly recommend you read the context before further testing.

dirkmueller:
+1

PR #219, opened 2015-05-11
Allow deploying neutron-server on SLE12 nodes

vuntz:
This seems to just work fine.

Also: Stop providing api-paste.ini as cookbook_file

This was only needed for the versions where the file could be different
from the packaged one, which was a long time ago. This should also not
be an issue on upgrades (content is correct, so packages will overwrite
the file now).


vuntz:
Smoketests passing in tempest; and 1525/1528 for full tests (with the 3 failures being the usual suspects). So looks pretty good.

jsuchome:
Seems to work well. +1

jsuchome:
Chaning to -1. After some time, controller node stopped responding. @mjura will file a bug report

vuntz:
> Changing to -1. After some time, controller node stopped responding. See

The bug mentions IP addresses not being assigned to the right interfaces, which would be a neutron-network issue, not a neutron-server issue.

jsuchome:
Hm, we're using SLE12 neutron-network for some time already, so I didn't suspect it. The example at crowbar.c13 cloud is still there, if you want to investigate

jsuchome:
(So, retracting my -1)

kbaikov:
+1

dirkmueller:
+1

PR #220, opened 2015-05-19
Only install metadata-agent where needed

rhafer:
The metadata-agent only needs to be installed on the compute nodes when using
DVR.


vuntz:
Need some testing, but +1 once tested.

rhafer:
I tested at least the non-DVR part. That looks good.

rsalevsky:
:+1:

PR #221, opened 2015-05-21
Set dhcp_agents_per_network option in neutron.conf

rsalevsky:
neutron-ha-tool replicates the dhcp only on start which is wrong an
causes new networks to be not replicated to the available dhcp-agent

this change effects that neutron keeps care about this behavior

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_config.rb
+++ b/chef/cookbooks/neutron/recipes/common_config.rb
@@ -120,6 +120,8 @@ if neutron[:neutron][:use_lbaas] then
   service_plugins = "#{service_plugins}, neutron.services.loadbalancer.plugin.LoadBalancerPlugin"
 end
 
+network_nodes = search_env_filtered(:node, "roles:neutron-network")
+
 template "/etc/neutron/neutron.conf" do
     cookbook "neutron"
     source "neutron.conf.erb"
@@ -150,7 +152,8 @@ template "/etc/neutron/neutron.conf" do
       :rootwrap_bin =>  node[:neutron][:rootwrap],
       :use_namespaces => true,
       :allow_overlapping_ips => neutron[:neutron][:allow_overlapping_ips],
-      :dvr_enabled => neutron[:neutron][:use_dvr]
+      :dvr_enabled => neutron[:neutron][:use_dvr],
+      :network_nodes_count => network_nodes.count
     }.merge(nova_notify))
 end
 
--- a/chef/cookbooks/neutron/templates/default/neutron.conf.erb
+++ b/chef/cookbooks/neutron/templates/default/neutron.conf.erb
@@ -191,7 +191,7 @@ allow_overlapping_ips = False
 
 # Number of DHCP agents scheduled to host a network. This enables redundant
 # DHCP agents for configured networks.
-# dhcp_agents_per_network = 1
+dhcp_agents_per_network = <%= @network_nodes_count %>
 
 # ===========  end of items for agent scheduler extension =====
 


vuntz:
Isn't there a command line argument in neutron-ha-tool for this?

rsalevsky:
Unfortunately no, here is the code for the tool: https://github.com/madkiss/openstack-resource-agents/blob/master/ocf/neutron-ha-tool

dguitarbite:
-1 This does not work.

On updating the number of network nodes (hence the services on network node), `$ neutron agent-list`  does not show the new services which it is supposed to do so. Also this does not update the number of dhcp agents in the neturon.conf file which is the work of this patch. I think we need to patch the HA neutron file upstream.

dguitarbite:
+1

rhafer:
> > Isn't there a command line argument in neutron-ha-tool for this?

> Unfortunately no, here is the code for the tool: https://github.com/madkiss/openstack-resource-agents/blob/master/ocf/neutron-ha-tool

I guess you're talking about different things here. neutron-ha-tool (as included in our neutron packages) has a switch to replicate all networks accross all available dhcp-agents. The bug was that the ocf-resource-agent for neutron-ha-tool only call neutron-ha-tool with that switch during the "start" action and not during "monitor". But since we originally created the resource-agent the new ```dhcp_agents_per_network``` option was added to neutron which is a lot cleaner IMO than relying on the external tool to distribute the networks. That why I suggested in the bug to use that option instead of changing the resource agent.

rhafer:
> On updating the number of network nodes (hence the services on network node), $ neutron agent-list does not show the new services which it is supposed to do so. I think we need to patch the HA neutron file upstream.

@dguitarbite I don't quite understand what you're saying here. Can you try to rephrase? What number did you update where?

chef/cookbooks/neutron/libraries/default.rb
```diff
@@ -0,0 +1,19 @@
+def is_crowbar?()
+  return defined?(Chef::Recipe::Barclamp) != nil
+end
+
+def get_network_nodes()
+  if is_crowbar?
```
rhafer: Do we really need this? I know we use this in cookbooks which are usable without Crowbar, but the neutron cookbook currently is not one of them. I am not strictly opposed to adding this here, but it seem weird to me and it certainly doesn't make the cookbook crowbar independent. :)

rhafer:
@rsalevsky Please add the ```bsc#928189```tag to the commit message's subject

rhafer:
@rsalevsky, @dguitarbite  Did you test what happens when creating a subnet when less than  ```dhcp_agents_per_network``` are running (e.g. because one of the network nodes is down)? Will network creation fail in that case?

dguitarbite:
@rhafer I just meant to say that I added another node to the HA cluster (network node). This failed to update the number of services as shown by $neutron agent-list. But later on Rick suggested that after waiting for some time, the service status are updated and the new services are displayed. I was counting the number of dhcp-agents, l3 agents etc.displayed by neutron agent-list

rsalevsky:
@rhafer I just tested it with @dguitarbite , works as expected.

rhafer:
> @rhafer I just tested it with @dguitarbite , works as expected.

@rsalevsky What means " works as expected"? :) Does creating networks / subnets still work if not enough agents are there?

rsalevsky:
Yes, creation of networks and subnets works.

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -150,7 +150,8 @@
       :rootwrap_bin =>  node[:neutron][:rootwrap],
       :use_namespaces => true,
       :allow_overlapping_ips => neutron[:neutron][:allow_overlapping_ips],
-      :dvr_enabled => neutron[:neutron][:use_dvr]
+      :dvr_enabled => neutron[:neutron][:use_dvr],
+      :network_nodes_count => get_network_nodes.count
```
vuntz: This should really just be: `search_env_filtered(:node, "roles:neutron-network").count`; no need to create a library for that.

rhafer:
The ```bsc``` tag is still missing in the subject. Sorry for being picky about this, but AFAIK ```obs-service-tar_scm``` only adds subject line to the changes file of the RPM and we like to have the bug ids to appear there.

chef/cookbooks/neutron/recipes/common_config.rb
rsalevsky: I tested this `search_env_filtered` is not available in this recipe/barclamp.I not sure why because neutron depend on `utils` where search_env_filtered is defined.

vuntz: Have you tried using it outside of the template? This should work there. So you just need to assign the results to a variable and use the variable here.

rsalevsky:
@rhafer Should be fine now.

rhafer:
+1

vuntz:
+1

PR #224, opened 2015-06-11
Allow switching the Keystone API version

rhafer:
Depends on: https://github.com/crowbar/barclamp-neutron/pull/197 and https://github.com/crowbar/barclamp-keystone/pull/280

Note there are currently a few places where the keystone v3 api is not supported by neutron (mostly neutron-ha-tool, still need to check what's breaking v3 for the metadata agent). In those place the v2 URL is still used for now.

### DIFF
--- a/chef/cookbooks/neutron/templates/default/metadata_agent.ini.erb
+++ b/chef/cookbooks/neutron/templates/default/metadata_agent.ini.erb
@@ -7,7 +7,10 @@ debug = <%= @debug ? "True" : "False" %>
 
 # The Neutron user information for accessing the Neutron API.
 # auth_url = http://localhost:5000/v2.0
-auth_url = <%= @keystone_settings['internal_auth_url'] %>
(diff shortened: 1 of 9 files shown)


dirkmueller:
+1

chef/cookbooks/neutron/templates/default/metadata_agent.ini.erb
```diff
@@ -7,7 +7,10 @@ debug = <%= @debug ? "True" : "False" %>
 
 # The Neutron user information for accessing the Neutron API.
 # auth_url = http://localhost:5000/v2.0
-auth_url = <%= @keystone_settings['internal_auth_url'] %>
+auth_url = <%= KeystoneHelper.versioned_service_URL(@keystone_settings["protocol"],
+                                                    @keystone_settings["internal_url_host"],
+                                                    @keystone_settings["service_port"],
+                                                    "2.0") %>
```
vuntz: Will we track these locations using v2.0 so we can fix them later on? Or we just don't care and will leave them that way forever?

PR #225, opened 2015-06-15
Set dhcp_agents_per_network only in HA mode (bsc#934651)

rsalevsky:


rhafer:
-1,  I don't think this is the right fix.

The problem is that the neutron-network role is only assigned to the network nodes after the neutron-server role has been deployed (at least when neutron-network and neutron-server are assigned to different clusters/nodes). So with the above fix the number would always be set to '1' of the first run of chef-client and only get corrected after the next run (about 15min later).

I guess the correct fix here is to get rid of the search and use the deployment attributes from the neutron proposal. Something like this (untested):

```network_nodes_count = node[:neutron][:elements][:neutron-network].count```

**Comment by rsalevsky** [comment]
Ah ok, now I see the real problem. I will test your proposed fix for the problem.

**Comment by rhafer** [comment]
Hm, actually  using ```node[:neutron][:elements][:neutron-network]``` is not enough. :( If ```neutron-network``` is deployed on a HA cluster it will just contain a single element ```cluster:<cluster-name>```.

The pacemaker barclamp however maintains a ```node[:neutron][:elements_expanded][:neutron-network]``` attribute which when deploying a HA setup contaions all the nodes that are part of that cluster. I am just not exactly sure currently if that attribute is already populated when deploying ```neutron-server```. BTW, ```elements_expanded``` attribute is not there, when deploying in non HA mode.

File: chef/cookbooks/neutron/recipes/common_config.rb [hunk 1]
```diff
@@ -120,7 +120,10 @@
   service_plugins = "#{service_plugins}, neutron.services.loadbalancer.plugin.LoadBalancerPlugin"
 end
 
-network_nodes = search_env_filtered(:node, "roles:neutron-network")
+network_nodes_count = 1
+if neutron[:neutron][:elements_expanded]
+  network_nodes_count = neutron[:neutron][:elements_expanded][:neutron-network].count
```
rhafer: Ruby doesn't seem to like ```-``` in symbols. I guess you need to use the string syntax here. :/

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -120,7 +120,10 @@
   service_plugins = "#{service_plugins}, neutron.services.loadbalancer.plugin.LoadBalancerPlugin"
 end
 
-network_nodes = search_env_filtered(:node, "roles:neutron-network")
+network_nodes_count = 1
+if neutron[:neutron][:elements_expanded]
+  network_nodes_count = neutron[:neutron][:elements_expanded]["neutron-network"].count
+end
```
mjura: Try to do this simpler:
```
network_nodes_count = neutron[:neutron][:elements_expanded]["neutron-network"].count if neutron[:neutron][:elements_expanded]
```

rsalevsky: This was my first version but the line is extremely long. So I prefer this syntax.

mjura:
LGTM +1

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -120,7 +120,10 @@
   service_plugins = "#{service_plugins}, neutron.services.loadbalancer.plugin.LoadBalancerPlugin"
 end
 
-network_nodes = search_env_filtered(:node, "roles:neutron-network")
+network_nodes_count = 1
```
vuntz: I'd prefer to initialize this with `neutron[:neutron][:elements]["neutron-network"].count` (which should be 1 anyway, but is more future-proof in case there are some changes later on).

rhafer:
+1

PR #227, opened 2015-06-16
Make "nova_floating" independent of "public" network

rhafer:


mjura:
Looks OK +1

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -276,7 +276,7 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
     allocate_virtual_ips_for_any_cluster_in_networks_and_sync_dns(server_elements, vip_networks)
 
     network_nodes.each do |n|
-      net_svc.allocate_ip "default", "public", "host",n
+      net_svc.enable_interface "default", "nova_floating", n
```
vuntz: We probably need something similar in the nova barclamp for the DVR case, I think.

rhafer: True I had that already. Just forgot to push it :(

rhafer:
> During testing I've got
> 
> ``` 2015-06-25 13:21:07.911 20400 INFO neutron.api.v2.resource [req-f5a82d71-0d19-43e9-9e1d-fc931e70a476 None] create failed (client error): Invalid input for operation: Gateway is not valid on subnet.```

@mjura Do you still have the network.json, that you used for this?

mjura:
> @mjura Do you still have the network.json, that you used for this?

No, I don't have. I had missing router attribute for separate `floating` network

crowbar_framework/app/models/neutron_service.rb
mjura: I've added missing PR 
https://github.com/crowbar/barclamp-nova/pull/469

mjura:
This will be merged on top of https://github.com/crowbar/barclamp-neutron/pull/230

rhafer:
> No, I don't have. I had missing router attribute for separate floating network

Actually this is not because for ```router``` missing in ```network_floating``` (there is a fallback to the router from ```public``` in that case). The problem is that the router (192.168.126.1) is not in the range of the floating network (192.168.126.128/25) anymore. So instead of adding a router to ```nova_floating``` in the network proposal I think we need some special handling here for the case where ```nova_floating``` is a subnet of ```public``` :(

mjura:
> The problem is that the router (192.168.126.1) is not in the range of the floating network (192.168.126.128/25) anymore. 

@rhafer : For my scenario I had different `floating` and `public` network, so there was no problem with router out of subnet range.

rhafer:
> @rhafer : For my scenario I had different floating and public network, so there was no problem with router out of subnet range.

For that case we have validation in network-json-validator now, that assures that there is a router in ```floating``` when it's not a subnet of ```public```.

rhafer:
Rebased on top of the multiple external networks change.

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -133,7 +133,7 @@
 
   # Create the bridges Neutron needs.
   # Usurp config as needed.
-  networks = [[ "nova_fixed", "fixed" ], ["public", "public"]]
+  networks = [[ "nova_fixed", "fixed" ], ["nova_floating", "public"]]
```
rsalevsky: `networks = [[ "nova_fixed", "fixed" ], [  "nova_floating", "public" ]]`

dirkmueller:
+1

rossella:
+1

PR #230, opened 2015-06-29
Enable setting multiple external networks

rossella:
Enable multiple external network for Neutron.
Some manual steps are still needed and will be documented in the
official documentation.

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_agent.rb
+++ b/chef/cookbooks/neutron/recipes/common_agent.rb
@@ -89,6 +89,7 @@ if neutron[:neutron][:networking_plugin] == 'vmware' or
   end
 end
 
+multiple_external_networks = !neutron[:neutron][:additional_external_networks].empty? && node.roles.include?("neutron-network")
 # openvswitch configuration specific to ML2
 if neutron[:neutron][:networking_plugin] == 'ml2' and
    neutron[:neutron][:ml2_mechanism_drivers].include?("openvswitch")
@@ -132,8 +133,11 @@ if neutron[:neutron][:networking_plugin] == 'ml2' and
 
   # Create the bridges Neutron needs.
   # Usurp config as needed.
-  [ [ "nova_fixed", "fixed" ],
-    [ "public", "public"] ].each do |net|
+  networks = [[ "nova_fixed", "fixed" ], ["public", "public"]]
+  neutron[:neutron][:additional_external_networks].each do |net|
+    networks << [net, net]
+  end
+  networks.each do |net|
     bound_if = (node[:crowbar_wall][:network][:nets][net[0]].last rescue nil)
     next unless bound_if
     name = "br-#{net[1]}"
@@ -185,7 +189,11 @@ else
   # Note: As moving between network plugins is currently not supported by this
   #       cookbook this code is mostly just sitting here and waiting for the
   #       plugin switching support to be implemented.
-  [ "br-public", "br-fixed" ].each do |name|
+  bridges = ["br-public", "br_fixed"]
+  neutron[:neutron][:additional_external_networks].each do |net|
+    bridges << "br-#{net}"
+  end
+  bridges.each do |name|
     service "ovs-usurp-config-#{name}" do
       # FIXME: Don't stop it here until we handle the shutdown of openvswitch
       #        and the neutron-ovs-agent correctly. Otherwise we might be cut
@@ -209,18 +217,34 @@ if neutron[:neutron][:networking_plugin] == "ml2"
   case
   when ml2_mech_drivers.include?("openvswitch")
     # package is already installed
-
     neutron_agent = node[:neutron][:platform][:ovs_agent_name]
     agent_config_path = "/etc/neutron/plugins/openvswitch/ovs_neutron_plugin.ini"
     interface_driver = "neutron.agent.linux.interface.OVSInterfaceDriver"
-    external_network_bridge = "br-public"
+    bridge_mappings = "floating:br-public"
+    if multiple_external_networks
+      bridge_mappings += ", "
+      bridge_mappings += neutron[:neutron][:additional_external_networks].collect { |n| n + ":" + "br-" + n }.join ','
+    end
+    if ml2_type_drivers.include?("vlan")
+      bridge_mappings += ", physnet1:br-fixed"
+    end
   when ml2_mech_drivers.include?("linuxbridge")
     package node[:neutron][:platform][:lb_agent_pkg]
 
     neutron_agent = node[:neutron][:platform][:lb_agent_name]
     agent_config_path = "/etc/neutron/plugins/linuxbridge/linuxbridge_conf.ini"
     interface_driver = "neutron.agent.linux.interface.BridgeInterfaceDriver"
-    external_network_bridge = ""
+    physnet = node[:crowbar_wall][:network][:nets][:nova_fixed].first rescue nil
+    interface_mappings = "physnet1:" + physnet
+    if multiple_external_networks
+      neutron[:neutron][:additional_external_networks].each do |net|
+        ext_iface = node[:crowbar_wall][:network][:nets][net].first
+        if ext_iface != physnet
+          mapping = ", " + net + ":" + ext_iface
+          interface_mappings += mapping
+        end
+      end
+    end
   end
 
   # include neutron::common_config only now, after we've installed packages
@@ -248,7 +272,8 @@ if neutron[:neutron][:networking_plugin] == "ml2"
         :ml2_type_drivers => ml2_type_drivers,
         :tunnel_types => ml2_type_drivers.select { |t| ["vxlan", "gre"].include?(t) },
         :use_l2pop => neutron[:neutron][:use_l2pop] && (ml2_type_drivers.include?("gre") || ml2_type_drivers.include?("vxlan")),
-        :dvr_enabled => neutron[:neutron][:use_dvr]
+        :dvr_enabled => neutron[:neutron][:use_dvr],
+        :bridge_mappings => bridge_mappings
       )
     end
   when ml2_mech_drivers.include?("linuxbridge")
@@ -268,11 +293,11 @@ if neutron[:neutron][:networking_plugin] == "ml2"
       group node[:neutron][:platform][:group]
       mode "0640"
       variables(
-        :physnet => (node[:crowbar_wall][:network][:nets][:nova_fixed].first rescue nil),
         :ml2_type_drivers => ml2_type_drivers,
         :vxlan_mcast_group => neutron[:neutron][:vxlan][:multicast_group],
-        :use_l2pop => neutron[:neutron][:use_l2pop] && ml2_type_drivers.include?("vxlan")
-      )
+        :use_l2pop => neutron[:neutron][:use_l2pop] && ml2_type_drivers.include?("vxlan"),
+        :interface_mappings => interface_mappings
+       )
     end
   end
 
@@ -302,7 +327,6 @@ if neutron[:neutron][:networking_plugin] == "ml2"
         :send_arp_for_ha => 3,
         :periodic_interval => 40,
         :periodic_fuzzy_delay => 5,
-        :external_network_bridge => external_network_bridge,
         :dvr_enabled => neutron[:neutron][:use_dvr],
         :dvr_mode => node.roles.include?("neutron-network") ? "dvr_snat" : "dvr"
       )
--- a/chef/cookbooks/neutron/recipes/common_agent.rb
+++ b/chef/cookbooks/neutron/recipes/common_agent.rb
@@ -189,7 +189,7 @@ else
   # Note: As moving between network plugins is currently not supported by this
   #       cookbook this code is mostly just sitting here and waiting for the
   #       plugin switching support to be implemented.
-  bridges = ["br-public", "br_fixed"]
+  bridges = ["br-public", "br-fixed"]
   neutron[:neutron][:additional_external_networks].each do |net|
     bridges << "br-#{net}"
   end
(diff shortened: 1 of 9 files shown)


rossella:
Tested with OVS, still need to test with Linuxbridge

rossella:
Tested with Linuxbridge too now

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -304,7 +338,7 @@
         :periodic_fuzzy_delay => 5,
         :external_network_bridge => external_network_bridge,
         :dvr_enabled => neutron[:neutron][:use_dvr],
-        :dvr_mode => node.roles.include?("neutron-network") ? "dvr_snat" : "dvr"
+        :dvr_mode => node.roles.include?("neutron-network") ? "dvr_snat" : "dvr",
```
vuntz: Unneeded change.

vuntz:
Interesting, need to look at it a bit deeper, but first thing I'm wondering is whether we should actually make the default floating network part of this attribute. This would then be the `external_networks` attribute, instead of `additional_external_networks`.

vuntz:
Does this require the additional networks to be defined in network.json? Or how does this work to get information on how to create these networks?

rossella:
@vuntz yes it requires the network to be added manually to network.json . You can find more info about the configuration needed here [1] (especially the last paragraph) .
It's possible to make the default floating network part of this attribute and I thought about it. It requires a previous cleanup since the name of the default floating network is hard coded in various part of the code (from network.json to barclamp neutron post_install_conf.rb) . 

[1] https://docs.google.com/document/d/1Q8Evv5Z6SViWhzEvZOme1xNlP4lv7Vd1owdoAuesxVI/edit?usp=sharing

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -89,6 +89,7 @@
   end
 end
 
+multiple_external_networks = !neutron[:neutron][:additional_external_networks].empty?  && node.roles.include?("neutron-network")
```
rsalevsky: two spaces between `.empty?  && node.`

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -132,8 +133,11 @@
 
   # Create the bridges Neutron needs.
   # Usurp config as needed.
-  [ [ "nova_fixed", "fixed" ],
-    [ "public", "public"] ].each do |net|
+  networks = [ [ "nova_fixed", "fixed" ], ["public", "public"]]
+  neutron[:neutron][:additional_external_networks].each do |net|
+    networks << [ net, net ]
```
rsalevsky: should be `[net, net]`

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -132,8 +133,11 @@
 
   # Create the bridges Neutron needs.
   # Usurp config as needed.
-  [ [ "nova_fixed", "fixed" ],
-    [ "public", "public"] ].each do |net|
+  networks = [ [ "nova_fixed", "fixed" ], ["public", "public"]]
```
rsalevsky: should `[[`

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -206,14 +214,27 @@
...
     interface_driver = "neutron.agent.linux.interface.OVSInterfaceDriver"
-    external_network_bridge = "br-public"
+    if !multiple_external_networks
+      external_network_bridge = "br-public"
+      if ml2_type_drivers.include?("vlan")
+        bridge_mappings = "physnet1:br-fixed"
+      end
+    else
+      bridge_mappings = "floating:br-public, "
+      bridge_mappings += neutron[:neutron][:additional_external_networks].collect {|n| n + ":" + "br-" + n }.join ','
```
rsalevsky: space missing after `{`

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -206,14 +214,27 @@
   ml2_mech_drivers = neutron[:neutron][:ml2_mechanism_drivers]
   ml2_type_drivers = neutron[:neutron][:ml2_type_drivers]
 
+
```
rsalevsky: empty line

mjura:
For me looks OK +1

I wondering if want to add these new external networks and create for them some additional routers and networks in neutron using `post_install_conf.rb` or should we leave it for user.

chef/data_bags/crowbar/bc-template-neutron.json
```diff
@@ -66,7 +67,7 @@
     "neutron": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 40,
+      "schema-revision": 41,
```
rhafer: Revision 41 was already taken by different pull request please take 42 (and rename 041_additional_external_networks.rb accordingly)

vuntz:
Shouldn't we integrate the `crowbar network enable_interface` bit from the doc in the rails app, when applying the barclamp?

rhafer:
> Shouldn't we integrate the crowbar network enable_interface bit from the doc in the rails app, when applying the barclamp?

Yeah. I'd like to see that as well. Should be pretty simple.

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -221,6 +242,18 @@
     agent_config_path = "/etc/neutron/plugins/linuxbridge/linuxbridge_conf.ini"
     interface_driver = "neutron.agent.linux.interface.BridgeInterfaceDriver"
     external_network_bridge = ""
+    physnet = node[:crowbar_wall][:network][:nets][:nova_fixed].first rescue nil
+    interface_mappings = "physnet1:" + physnet
+    if multiple_external_networks
+      neutron[:neutron][:additional_external_networks].each do |net|
+        n = ":"+ net
```
rhafer: What is this for? ```n``` doesn't seem to be used anywhere.

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -206,14 +214,27 @@
   ml2_mech_drivers = neutron[:neutron][:ml2_mechanism_drivers]
   ml2_type_drivers = neutron[:neutron][:ml2_type_drivers]
 
+
   case
   when ml2_mech_drivers.include?("openvswitch")
-    # package is already installed
 
+    # package is already installed
```
rhafer: Nitpick: Unneeded change.

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -206,14 +214,27 @@
...
     neutron_agent = node[:neutron][:platform][:ovs_agent_name]
     agent_config_path = "/etc/neutron/plugins/openvswitch/ovs_neutron_plugin.ini"
     interface_driver = "neutron.agent.linux.interface.OVSInterfaceDriver"
-    external_network_bridge = "br-public"
+    if !multiple_external_networks
```
rhafer: Do we really need to distinguish the multiple vs single external network cases here? Can't we always use the code path that's now in the ```else``` branch?

rhafer:
> > Shouldn't we integrate the crowbar network enable_interface bit from the doc in the rails app, when applying the barclamp?
> 
> Yeah. I'd like to see that as well. Should be pretty simple. (But it would require a similar change to the nova barclamp as well)


@rossella  I just submitted a pull request to your branch that should do the trick.

chef/cookbooks/neutron/recipes/common_agent.rb
rossella: I can refactor a bit, you are right. I still need and if/else block because external_network_bridge should not be set if multiple external networks are used

chef/cookbooks/neutron/recipes/common_agent.rb
rossella: sorry, leftover from a previous version

rossella:
@rhafer thanks  a lot!

rossella:
@mjura some manual steps are still required to set up multiple networks so I think it's OK to leave the task of creating the router and the external network to the user for now.

vuntz:
> @rossella I just submitted a pull request to your branch that should do the trick.

For reference, https://github.com/rossella/barclamp-neutron/pull/1

> @mjura some manual steps are still required to set up multiple networks

@rossella which manual steps are still required? I think I'd like to see the openstack resources created automatically also (but could be in a later PR, of course).

(This PR will be a great change, btw!)

chef/cookbooks/neutron/recipes/common_agent.rb
rhafer: Is ```external_network_bridge``` really required in the case when there is just a single external network? Or can we do without it in that case as well?

rossella: As long as the bridge_mappings is set it's fine if external_network_bridge is not set...I removed external_network_bridge completely and set the bridge_mappings all the time

rossella:
@vuntz the manual step required is to modify /etc/crowbar/network.json to add the additional external networks. Is there an easy way to automate that?

rossella:
@mjura @vuntz I'd prefer to automate the creation of the OpenStack resources in another PR...the post install script needs some refactoring and this PR is getting quite big already

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -185,7 +189,11 @@
   # Note: As moving between network plugins is currently not supported by this
   #       cookbook this code is mostly just sitting here and waiting for the
   #       plugin switching support to be implemented.
-  [ "br-public", "br-fixed" ].each do |name|
+  bridges = ["br-public", "br_fixed"]
```
dirkmueller: why did you change from br-fixed to br_fixed ?

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -209,18 +217,34 @@
...
+    end
   when ml2_mech_drivers.include?("linuxbridge")
     package node[:neutron][:platform][:lb_agent_pkg]
 
     neutron_agent = node[:neutron][:platform][:lb_agent_name]
     agent_config_path = "/etc/neutron/plugins/linuxbridge/linuxbridge_conf.ini"
     interface_driver = "neutron.agent.linux.interface.BridgeInterfaceDriver"
-    external_network_bridge = ""
+    physnet = node[:crowbar_wall][:network][:nets][:nova_fixed].first rescue nil
+    interface_mappings = "physnet1:" + physnet
```
dirkmueller: the rescue nil and then +physnet in the line below doesn't look good to me. if this really can not happen, then remove the rescue nil (avoid hiding the real error). if it can happen, the rescue should return something more sensible.

chef/cookbooks/neutron/templates/default/l3_agent.ini.erb
```diff
@@ -50,7 +50,6 @@ handle_internal_only_routers = <%= @handle_internal_only_routers %>
 # empty value for the linux bridge. when this parameter is set, each L3 agent
 # can be associated with no more than one external network.
 # external_network_bridge = br-ex
-external_network_bridge = <%= @external_network_bridge %>
```
dirkmueller: hmm, why does the default value of 'br-ex' work?

chef/cookbooks/neutron/templates/default/linuxbridge_conf.ini.erb
```diff
@@ -34,7 +34,7 @@
 #
 # physical_interface_mappings =
 # Example: physical_interface_mappings = physnet1:eth1
-physical_interface_mappings = physnet1:<%=@physnet%>
+physical_interface_mappings = <%=@interface_mappings%>
```
dirkmueller: suggest additional spacing, e.g.

```ruby
   <%= @interface_mappings -%>
```

chef/cookbooks/neutron/templates/default/l3_agent.ini.erb
rossella: I don't think it works, it's commented...am I missing something?

chef/cookbooks/neutron/recipes/common_agent.rb
rossella: ooops

chef/cookbooks/neutron/recipes/common_agent.rb
rossella: rescue nil removed

chef/cookbooks/neutron/templates/default/l3_agent.ini.erb
mjura: For me it is ok look comment above `when this parameter is set, each L3 agent can be associated with no more than one external network`

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -277,6 +299,9 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
 
     network_nodes.each do |n|
       net_svc.allocate_ip "default", "public", "host",n
```
mjura: `net_svc.allocate_ip "default", "public", "host",n`

can be removed, look on comment from:
https://github.com/crowbar/barclamp-neutron/pull/227/files#diff-1cab7f3b6842279ae4d6bd875e073cacL279

chef/cookbooks/neutron/templates/default/l3_agent.ini.erb
dirkmueller: then we should imho explicitely set external_network_bridge to empty

crowbar_framework/app/models/neutron_service.rb
rhafer: > net_svc.allocate_ip "default", "public", "host",n
> 
> can be removed,

That's not really true. It could be changed to an appropriate enable_interface call. But let's worry about that in the other pull request (#227).

mjura:
+1 if you will fix comment from @dirkmueller

dirkmueller:
LGTM

vuntz:
For the record: this seems to break ovs support.

PR #231, opened 2015-06-30
Introduce HoundCI

rsalevsky:
We want to test HoundCI, this configs are needed to run Hound
with our preferred styles.

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,0 +1,2 @@
+ruby:
+  config_file: .rubocop.yml
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,23 @@
+inherit_from: .rubocop_todo.yml
+
+AllCops:
+  RunRailsCops: true
+  DisplayCopNames: true
+
+Style/StringLiterals:
+  Enabled: true
+  EnforcedStyle: double_quotes
+
+Style/StringLiteralsInInterpolation:
+  Enabled: true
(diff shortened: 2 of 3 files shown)


dirkmueller:
@rsalevsky was there any discussion on the preferred style anywhere?

I like consistency though, so in general I'm in favor. is there a way to see a complete report with the settings that you're proposing to see how bad it is compared to the current code base?

rsalevsky:
We use the default ruby style with some adaptions which we use already in our codebase. I crated a full [rubocop log](https://gist.githubusercontent.com/rsalevsky/656bc135858763faca35/raw/28194609cf8266be49502293e686135735836cb3/rubocop) with all detected offenses.

In short: 48 files inspected, 1285 offenses detected

dirkmueller:
looks like top 5 is:

    409  Style/StringLiterals
    212  Style/HashSyntax
    196  Metrics/LineLength
     60  Style/SpaceInsideBrackets
     40  Style/MethodDefParentheses


since StringLiterals is just a matter of taste and HashSyntax actually makes backporting to Cloud 4 more difficult I'd prefer to disable those. LineLenght is debateable (I personally actually like the smaller length). 

Anyway, this should be the same in all crowbar barclamps and not per barclamp

tboerger:
Let's close this for now and let's keep this initially only on nova and crowbar to do some refinement of the rules. When we all agree on the temporary style we can start to distribute it accross the barclamps.

PR #236, opened 2015-07-10
Update the interface in ovs-usurp-config service

aplanas:
Temporal workaround for bsc#930986 proposed by rhafer.
A better workaround based on
https://en.opensuse.org/Portal:Wicked/OpenvSwitch is in
planning.

### DIFF
--- a/chef/cookbooks/neutron/templates/default/ovs-usurp-config.erb
+++ b/chef/cookbooks/neutron/templates/default/ovs-usurp-config.erb
@@ -21,6 +21,11 @@ case "$1" in
         ip -4 route flush dev <%= @source %>
         ip -6 route flush dev <%= @source %>
         ip addr flush dev <%= @source %>
+        # (aplanas) temporal workaround for bsc#930986 proposed by
+        # rhafer. A better workaround based on
+        # https://en.opensuse.org/Portal:Wicked/OpenvSwitch is in
+        # planning.
+        ip link set up <%= @source %>
 <% @addresses.each do |addr| -%>
         ip addr add <%= addr.to_s %> dev <%= @dest %>
 <% end -%>


vuntz:
I don't fully understand why it's needed now (probably wicked), but can't see why it would hurt, so +1.

aplanas:
Is indeed the interaction between Wicked, OVS and Systemd.  In the future is expected that Wicked will create the bridge, making both workarounds (the one from this PR and the one described in the wiki) outdated.

chef/cookbooks/neutron/templates/default/ovs-usurp-config.erb
```diff
@@ -18,6 +18,11 @@ rc_reset
 
 case "$1" in
     start)
+        # (aplanas) temporal workaround for bsc#930986 proposed by
+        # rhafer. A better workaround based on
+        # https://en.opensuse.org/Portal:Wicked/OpenvSwitch is in
+        # planning.
+        ip link set up <%= @source %>
```
rhafer: Does this really work here? I expected it to be needed after the flush calls. Or even as the very last step of "start"

rhafer:
@aplanas did you create a follow up cards for the ifcfg based approach in trello?

chef/cookbooks/neutron/templates/default/ovs-usurp-config.erb
aplanas: Worked in both places, as far as I tested. For not reason I decided to put here at the beginning.  I will move this bellow just in case.

aplanas:
@rhafer Is documented here https://trello.com/c/EWy13eix/

aplanas:
* Rebase
* Move the `ip link set up` after `ip addr flush`

rhafer:
+1

@aplanas please cherry-pick into tex

PR #241, opened 2015-07-23
Do not add bridge mappings for external networks on compute nodes

vuntz:
The compute nodes don't need that, unless DVR is used.


vuntz:
@cc rossella

dirkmueller:
LGTM

rossella:
good catch! I had that check at some point then I must have removed it by mistake in the last versions of the multiple external network patch

+1

PR #242, opened 2015-07-27
Create fwaas / lbaas tables in the database

vuntz:
This is required since the split of these components out of neutron.


dirkmueller:
I wonder if that should be fixed in the packages (as well)?

vuntz:
> I wonder if that should be fixed in the packages (as well)?

That's a good point. I'm not totally sure how that would be done; should we check if the fwaas/lbaas package is installed and do that in the neutron-server init script?

dirkmueller:
+1

PR #243, opened 2015-07-28
Do not run multiple api/rpc processes

dirkmueller:
This requires excessive amount of memory and causes the mkcloud
runs to fail randomly. In addition this is marked as experimental
upstream and currently not well tested.

### DIFF
--- a/chef/cookbooks/neutron/templates/default/neutron.conf.erb
+++ b/chef/cookbooks/neutron/templates/default/neutron.conf.erb
@@ -216,11 +216,7 @@ dhcp_agents_per_network = <%= @network_nodes_count %>
 # worker thread in the current process.  Greater than 0 launches that number of
 # child processes as workers.  The parent process manages them.
 # api_workers = 0
-# This is workaround for lp#1354226
-# https://bugs.launchpad.net/neutron/+bug/1354226
-<% unless @core_plugin == "vmware" -%>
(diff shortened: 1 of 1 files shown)


chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -216,11 +216,7 @@ dhcp_agents_per_network = <%= @network_nodes_count %>
 # worker thread in the current process.  Greater than 0 launches that number of
 # child processes as workers.  The parent process manages them.
 # api_workers = 0
-# This is workaround for lp#1354226
-# https://bugs.launchpad.net/neutron/+bug/1354226
-<% unless @core_plugin == "vmware" -%>
-api_workers = <%= [node["cpu"]["total"], 2, 4].sort[1] %>
-<% end -%>
+api_workers = 0
```
vuntz: Is this one also experimental?

vuntz:
Interesting; I guess that would remove the need for having 6GB of memory for the controller that I noticed with the switch to kilo...

chef/cookbooks/neutron/templates/default/neutron.conf.erb
dirkmueller: non zero api workers is not experimental, but its helpful in removing the memory overhead..

toabctl:
+1 (and it's already in the rpm package)

PR #244, opened 2015-07-30
Fix search for nova instance

dirkmueller:
When nova-multi-controller and neutron-server are deployed
on two different nodes, the neutron.conf was not properly updated
with the nova settings because it could't find the node. Search
for the instance (optional since normal deployment order currently
is neutron first, then nova) properly so that get_instance()
in the chef recipe can do its job.

(cherry picked from commit fb1cea50ca2467e8a3738fac201d74ead38053f9)


toabctl:
+1

toabctl:
This is already since ages in the rpm package and well tested. Merging...

PR #245, opened 2015-08-03
Write network configuration to have wicked understand ovs (bsc#930986)

vuntz:
This is based on https://en.opensuse.org/Portal:Wicked/OpenvSwitch and
comment 9 in the bug.

Depends on https://github.com/crowbar/barclamp-network/pull/327

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_agent.rb
+++ b/chef/cookbooks/neutron/recipes/common_agent.rb
@@ -133,6 +133,8 @@ if neutron[:neutron][:networking_plugin] == 'ml2' and
 
   # Create the bridges Neutron needs.
   # Usurp config as needed.
+  has_ovs_bridge = false
+
   networks = [[ "nova_fixed", "fixed" ], [ "nova_floating", "public" ]]
   neutron[:neutron][:additional_external_networks].each do |net|
     networks << [net, net]
@@ -140,6 +142,9 @@ if neutron[:neutron][:networking_plugin] == 'ml2' and
   networks.each do |net|
     bound_if = (node[:crowbar_wall][:network][:nets][net[0]].last rescue nil)
     next unless bound_if
+
+    has_ovs_bridge = true
+
     name = "br-#{net[1]}"
     execute "Neutron: create #{name}" do
       command "ovs-vsctl add-br #{name}; ip link set #{name} up"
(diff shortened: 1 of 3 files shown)


vuntz:
cc @rhafer since he might have an opinion on the uglyness of the approach ;-)

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -157,53 +162,140 @@
...
+
+    if node[:platform] == "suse" && node[:platform_version].to_f >= 12.0
+      # on SLE12, we create proper network configuration for OVS bridges, see
+      # https://en.opensuse.org/Portal:Wicked/OpenvSwitch
+
+      # We need to create a static ifcfg file so that wicked will know that the
+      # ovs bridge usurps the interface address. This also means that we need
+      # to override the variables we pass to the template for the ifcfg of the
+      # interface.
+      bound_if_resource = resources(:template => "/etc/sysconfig/network/ifcfg-#{bound_if}")
```
aplanas: Isn't the creation and management of bridges and interfaces a responsibility of barclamp-network?

I was testing that if I check the bridge option for in interface in the node information, and I include the recipe from network, this will update (wrongly) the interfaces.

IIUC one of the risks of managing the interfaces here is that barclamp-network can undo this.

vuntz: > Isn't the creation and management of bridges and interfaces a responsibility of barclamp-network?

Yes and no. The issue is that the network cookbook doesn't have information about what is needed here, and that's information internal to the neutron cookbook.

> I was testing that if I check the bridge option for in interface in the node information, and I include the recipe from network, this will update (wrongly) the interfaces.

Hrm, not sure I understand. Can you rephrase or give more details?

> IIUC one of the risks of managing the interfaces here is that barclamp-network can undo this.

When would this happen? As long as the neutron cookbook is being used, then it will not be undone.

aplanas:
+1 Tested with wicked updated

chef/cookbooks/neutron/recipes/common_agent.rb
dirkmueller: I wonder if this could be moved to barclamp-network instead. it shouldn't be impossible to figure out that we potentially run neutron on that node.. just need to search for an active proposal and then deploy the usurp part there..

rsalevsky:
@vuntz, this repository got merged into [crowbar-openstack](https://github.com/crowbar/crowbar-openstack).
Please open a new pull request there. __Do NOT merge here anymore!__

rsalevsky:
Moved to https://github.com/crowbar/barclamp-neutron/pull/245

PR #246, opened 2015-08-06
Disable multiple external network support

toabctl:
There seems to be a problem and we have revert-external-network.patch
in our rpm package to revert the support.
So multiple external network support needs some rework and we disable
it for now.


dirkmueller:
+1, since its already in the packages..

PR #248, opened 2015-08-07
Be conservative with starting metadata agents

dirkmueller:
The default behavior of starting #cpu/2 "extra" agents makes
more sense, but still give it a top of 4 metadata agents.


toabctl:
+1

PR #249, opened 2015-08-10
Write network configuration to have wicked understand ovs (bsc#930986)

aplanas:
This is based on https://en.opensuse.org/Portal:Wicked/OpenvSwitch and
comment 9 in the bug.

(cherry picked from commit e6d3d94947e1528724750e3dbc9fc53c5df8356a)

### DIFF
--- a/chef/cookbooks/neutron/files/suse/ifcfg-ovs-system
+++ b/chef/cookbooks/neutron/files/suse/ifcfg-ovs-system
@@ -1,0 +1,5 @@
+# Managed by Crowbar.
+# Do not edit.
+name='ovs-system'
+STARTMODE=hotplug
+BOOTPROTO=none
--- a/chef/cookbooks/neutron/recipes/common_agent.rb
+++ b/chef/cookbooks/neutron/recipes/common_agent.rb
@@ -167,10 +167,15 @@ if neutron[:neutron][:networking_plugin] == 'ml2' and
 
   # Create the bridges Neutron needs.
   # Usurp config as needed.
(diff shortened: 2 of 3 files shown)


ovisan:
+1

rhafer:
According to https://github.com/crowbar/crowbar-openstack/pull/3#issuecomment-137975166 the wicked restart that this PR does breaks HA setups.

Also I am wondering if we really need to backport the fix to tex. We have a workaround already merged (https://github.com/crowbar/barclamp-neutron/commit/daf40ea008bf03f1dd1996e6895c85585d6185c9) that might be enough for tex. After all, this patch (or the alternative being implemented in https://github.com/crowbar/crowbar-core/pull/33) changes quite a bit of complex code. At least I would rather see it stable and proven to work in "master" before attempting a backport.

@aplanas, @vuntz does that make sense?

aplanas:
@rhafer for me makes sense to continue testing it only in suse cloud 6.  But did you try a HUP to wicked instead of a full restart?

vuntz:
> Also I am wondering if we really need to backport the fix to tex. We have a workaround already merged (daf40ea) that might be enough for tex

Actually, afaik, the workaround was not enough. Or am I misremembering?

rhafer:
Ugh. I thought it was. But actually we might need the workaround for https://bugzilla.suse.com/show_bug.cgi?id=945219 as well.

vuntz:
@rhafer should we kill this PR (since I believe you're working on this)?

rhafer:
@vuntz yes please.

PR #250, opened 2015-08-12
Install fwaas packages on the neutron-network role node

toabctl:
The l3-agent needs the fwaas package. Otherwise you get:

    ERROR neutron.services.firewall.agents.l3reference.firewall_l3_agent \
          FWaaS plugin is configured in the server side,
          but FWaaS is disabled in L3-agent.

and the l3 agent service can not start.


dirkmueller:
LGTM

rsalevsky:
+1

PR #251, opened 2015-08-12
Set proper network_type for the floating network

dirkmueller:
The floating network is never a GRE type network, its either
VLAN or flat.

### DIFF
--- a/chef/cookbooks/neutron/recipes/post_install_conf.rb
+++ b/chef/cookbooks/neutron/recipes/post_install_conf.rb
@@ -74,20 +74,16 @@ networking_plugin = node[:neutron][:networking_plugin]
 ml2_type_drivers_default_provider_network = node[:neutron][:ml2_type_drivers_default_provider_network]
 case networking_plugin
 when 'ml2'
+
+  floating_network_type = "--provider:network_type flat --provider:physical_network floating"
+
   case ml2_type_drivers_default_provider_network
   when 'vlan'
     fixed_network_type = "--provider:network_type vlan --provider:segmentation_id #{fixed_net["vlan"]} --provider:physical_network physnet1"
-    if node[:network][:networks][:nova_floating][:use_vlan]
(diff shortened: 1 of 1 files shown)


rossella:
LGTM

chef/cookbooks/neutron/recipes/post_install_conf.rb
```diff
@@ -84,7 +84,11 @@ def mask_to_bits(mask)
     end
   when 'gre'
     fixed_network_type = "--provider:network_type gre --provider:segmentation_id 1"
-    floating_network_type = "--provider:network_type gre --provider:segmentation_id 2"
+    if node[:network][:networks][:nova_floating][:use_vlan]
```
rossella: maybe this should be moved after line 78? We should take care of the floating network just once not for every type (gre, vxlan, etc) so we can remove some lines like 81-83 etc

dirkmueller: yeah, I was wondering about that- I'm not sure about the vxlan case (see below).

toabctl: > yeah, I was wondering about that- I'm not sure about the vxlan case (see below).

For provider networks, I think only `flat` and `vlan` are useful. From http://docs.openstack.org/admin-guide-cloud/content/tenant-provider-networks.html : 
```
Provider networks. Provider networks are created by the OpenStack administrator and map directly to an existing physical network in the data center. Useful network types in this category are flat (untagged) and VLAN (802.1Q tagged). It is possible to allow provider networks to be shared among tenants as part of the network creation process.
```

dirkmueller: we can try the other variant as well..

rossella:
LGTM now

rsalevsky:
LGTM

PR #252, opened 2015-08-13
Revert "Disable multiple external network support"

dirkmueller:
With the underlying regression fixed, we can reenable the
support for multiple external networks.

This reverts commit 18e8e32ab23ce0c112bc281594718f8d6a987cb2.


dirkmueller:
It seems gating passes so it should be obvious to merge this..

PR #257, opened 2015-09-07
[WIP] Allow to set use_external_dhcp

rossella:
This patch makes possible to set the flag use_external_dhcp.
If use_external_dhcp is true a metadata proxy will be spawned by
the DHCP agent even if the subnet has dhcp_enable set to False.
This is useful for cloud where an external appliance is used to
provide DHCP services.


rossella:
This is a WIP ...please don't merge. I just wanted to share the code and get some suggestions. The neutron patch that is needed to make this work is here https://github.com/rossella/neutron/commit/f6cc6012f6f919378d68c10df220858c22afe350

vuntz:
+1 for when this will be okay to merge.

@rossella I'm adding [WIP] to the title of the PR for now, please remove it once it's not WIP anymore :-)

PR #258, opened 2015-09-24
Create "floating" network as "flat" provider network (bsc#946882)

rhafer:
This should work regardless of the mechanism driver that is used  (openvswitch
or linuxbridge) as we have the ml2 type driver "flat" enabled unconditionally.
This is also in line with the behaviour of devstack.

For linuxbridge setups this commit also adjust the interface mappings for
external networks to use the VLAN interface instead for the underlying physical
interface. That's required for "flat" networks.

We also make sure to specify the right "--provider:physical_network" parameter
for the floating network. Otherwise the floating network might endup on the
wrong phyiscal interfaces in deployments that use different network conduits
for "nova_floating" and "nova_fixed".

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_agent.rb
+++ b/chef/cookbooks/neutron/recipes/common_agent.rb
@@ -256,6 +256,12 @@ if neutron[:neutron][:networking_plugin] == "ml2"
     agent_config_path = "/etc/neutron/plugins/linuxbridge/linuxbridge_conf.ini"
     interface_driver = "neutron.agent.linux.interface.BridgeInterfaceDriver"
     external_network_bridge = ""
+    physnet = node[:crowbar_wall][:network][:nets][:nova_fixed].first
+    interface_mappings = "physnet1:" + physnet
+    if neutron[:neutron][:use_dvr] || node.roles.include?("neutron-network")
+      floatingphys = node[:crowbar_wall][:network][:nets][:nova_floating].last
+      interface_mappings += ", floating:" + floatingphys
+    end
   end
 
   # include neutron::common_config only now, after we've installed packages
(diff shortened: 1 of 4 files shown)


rhafer:
tex backport of https://github.com/crowbar/crowbar-openstack/pull/44

mjura:
+1

vuntz:
Won't that break existing deployments where the networks are already created?

vuntz:
(code-wise, +1, but really unsure about my question)

rhafer:
> Won't that break existing deployments where the networks are already created?

Hm, good question, I didn't test that. Will do.

But I am pretty sure it won't break existing systems. The networks are not recreated, the physnet `floating` wasn't used anywhere before and the interface mapping for `physnet1` does not change. Did I miss something?

vuntz:
@rhafer I'm mostly worried about the config change `flat_networks = floating`

rhafer:
@vuntz You know that that has nothing to do with the neutron network/subnetwork `floating`? It's just a name for the `physnet` that is later mapped to `eth0.300` (by default at least). It could also be called  `physnet2` (or something else completely arbitrary).

But as said. I am running an extra round of testing for the update case.

vuntz:
> @vuntz You know that that has nothing to do with the neutron network/subnetwork floating?

No, didn't know :-)

rhafer:
I can confirm that this does not break existing deployments. I deployed this patch in an existing linuxbridge enviroment (ran chef-client on all nodes after that) and I could still reach floating IPs on instances that already ran before the update and successfully launch new instances and attache floating IPs.

@vuntz Should be fine to remove the "do not merge" label again.

PR #259, opened 2015-09-25
Replug vlan in OVS after enslaving parent device

rhafer:
This is a workaround for bnc#945219 to avoid network connectivity issue.
Backported from: 486c4cff7481f171febc7ae373c33a052152fa80 (crowbar-core)
Also adapted the bridge creation code to use the backported enhancements of the
OvsBridge Class.

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_agent.rb
+++ b/chef/cookbooks/neutron/recipes/common_agent.rb
@@ -172,21 +172,33 @@ if neutron[:neutron][:networking_plugin] == 'ml2' and
     bound_if = (node[:crowbar_wall][:network][:nets][net[0]].last rescue nil)
     next unless bound_if
     name = "br-#{net[1]}"
-    execute "Neutron: create #{name}" do
-      command "ovs-vsctl add-br #{name}; ip link set #{name} up"
-      not_if "ovs-vsctl list-br |grep -q #{name}"
-    end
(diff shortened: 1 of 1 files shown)


rhafer:
This depends on: 
https://github.com/crowbar/barclamp-deployer/pull/252

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -172,21 +172,33 @@
...
-    execute "Neutron: create #{name}" do
-      command "ovs-vsctl add-br #{name}; ip link set #{name} up"
-      not_if "ovs-vsctl list-br |grep -q #{name}"
-    end
-    execute "Neutron: add #{bound_if} to #{name}" do
-      command "ovs-vsctl del-port #{name} #{bound_if} ; ovs-vsctl add-port #{name} #{bound_if}"
-      not_if "ovs-dpctl show system@#{name} | grep -q #{bound_if}"
+    ruby_block "Neutron: create #{name}" do
+      block do
+          Nic::OvsBridge.create(name)
```
rhafer, line 177: If you think switch to use Nic::OvsBridge here is to risky a change we could also stay here with calling ovs directly. But I think this makes the code (a little bit) cleaner.

rhafer:
Closing this as this did not really work. Currently working on a different approach.

PR #261, opened 2015-11-12
Fix migration of neutron-l3 to neutron-network in HA case (bsc#954735)

vuntz:
We also need to migrate the elements_expanded attribute.

https://bugzilla.suse.com/show_bug.cgi?id=954735

### DIFF
--- a/chef/data_bags/crowbar/migrate/neutron/030_fix_rename_l3_role.rb
+++ b/chef/data_bags/crowbar/migrate/neutron/030_fix_rename_l3_role.rb
@@ -1,0 +1,17 @@
+def upgrade ta, td, a, d
+  if d.fetch('elements_expanded', {}).has_key? 'neutron-l3'
+    d['elements_expanded']['neutron-network'] = d['elements_expanded']['neutron-l3']
+    d['elements_expanded'].delete('neutron-l3')
+  end
+
+  return a, d
+end
+
+def downgrade ta, td, a, d
+  if d.fetch('elements_expanded', {}).has_key? 'neutron-network'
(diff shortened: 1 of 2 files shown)


rhafer:
Looks good to me. The bug report also mentions that the old role (neutron-l3) not deleted from chef. Any plans to add that here as well?

vuntz:
I don't think the old role matters that much, tbh, and I'm not convinced we should deal with that in migrations.

vuntz:
Pressed "comment" too fast (second time this morning...)... Is there any big issue with having the old role still out there?

rhafer:
I don't think there is any big issue with keeping it. It would just avoid any confusion in the futrure. But I agree that the migration might be the wrong place. It's probably better handled during a major release upgrade (tex -> tex+1). We just need to make sure to not forget about that.

vuntz:
@rhafer this is already tracked in https://trello.com/c/mCWliZfA/485-cloud6-barclamp-neutron-cleanup actually (which is tracked in an epic for cloud 6)

rhafer:
Ok. +1

toabctl:
+1

chef/data_bags/crowbar/migrate/neutron/030_fix_rename_l3_role.rb
```diff
@@ -0,0 +1,17 @@
+def upgrade ta, td, a, d
+  if d.fetch('elements_expanded', {}).has_key? 'neutron-l3'
```
tboerger, line 2: `if d.fetch('elements_expanded', {}).key? 'neutron-l3'` that's what hound would suggest.

tboerger:
A single comment, otherwise :+1:

PR #260, opened 2015-10-15
Backport ovs related additions from crowbar-openstack

rhafer:
This depends on: https://github.com/crowbar/barclamp-network/pull/331

PR #262, opened 2015-11-23
Do not rely on notification to enable ovs-usurp-config scripts

vuntz:
With a notification, if chef-client crashes before the end, the script
will never be enabled. Just always enable it, there's no reason why we
would want to not do this in the case where we have the script.

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_agent.rb
+++ b/chef/cookbooks/neutron/recipes/common_agent.rb
@@ -211,8 +211,9 @@ if neutron[:neutron][:networking_plugin] == 'ml2' and
     end
     service "ovs-usurp-config-#{name}" do
       # Don't start it here. It only needs to be executed during boot.
-      action [:nothing]
-      subscribes :enable, resources("template[/etc/init.d/ovs-usurp-config-#{name}]")
+      action [:enable]
+      # see comment in template above
+      not_if { addresses.empty? and routes.empty? }
     end
   end
 else


vuntz:
While we wait for https://github.com/crowbar/barclamp-neutron/pull/260...

@rhafer: I'm unsure why it was done in a notification, but maybe I'm missing something.

rhafer:
@vuntz IIRC I did it with a notification so that it is only enabled when the script was really generated. There are some types of setup where br-fixed and br-public are created but don't need to have any IP address assigned from the underlying interface. E.g. in a "dual" setup when "nova_fixed" or "nova_public" are assigned to eth1/eth1.300.
As an alternative to the notfication it shoud work to use the same `not_if` connection as for the `template` (though that might fail after a crash as well). Or an `only_if` to check of the existance of the init script.

vuntz:
@rhafer thanks for the background. I've added the `not_if`. Now not sure if it's worth merging or if we should just wait for your other PR. I'll leave that decision to you ;-)

rhafer:
@vuntz Hm, does the `not_if` really work? I somehow think that it fails in the same way as the notification when chef-client crashes. 

At least if it crashes after the `ruby_block "Have #{name} usurp config from #{bound_if}" do` and before the `service "ovs-usurp-config-#{name}"`.  If the risk of having chef crash at that point is small enough, shouldn't it be just enough to just change the notification to:

```
subscribes :enable, resources("template[/etc/init.d/ovs-usurp-config-#{name}]"), :immediately
```

?

vuntz:
> @vuntz Hm, does the not_if really work? I somehow think that it fails in the same way as the notification when chef-client crashes.

Hrm, why? It's the next instruction after the template, so there should be no crash between the template and this. I'm probably misunderstanding something in your comment :-)

Generally speaking, though: I prefer to have an explicit `:enable` instead of relying on a notification because then we enforce that the service is enabled on each chef run.

rhafer:
> Generally speaking, though: I prefer to have an explicit :enable instead of relying on a notification because then we enforce that the service is enabled on each chef run.

This case is somewhat special though. For subsequent test runs the `not_if` condition would prevent the enablement of the service (See: https://github.com/crowbar/barclamp-neutron/pull/262/files#diff-91474df28ca9d038c83a65ad4c220591R206 for details).

vuntz:
@rhafer only if ovs-usurp-config-foo has run on the reboot; if it hasn't run but the init script exists (ie: init script was created but is not enabled), then that should run.

I know, it's kind of a edge case, and again, maybe not worth investing efforts into it because of your other PR. I just happened to have hit this.

rhafer:
> only if ovs-usurp-config-foo has run on the reboot; if it hasn't run but the init script exists (ie: init script was created but is not enabled), then that should run.

Hm, good point.

> I know, it's kind of a edge case, and again, maybe not worth investing efforts into it because of your other PR. I just happened to have hit this.

Too many the edge cases lately ;)

That being said I am fine with this: +1. But it should at least pass ci I guess.

vuntz:
> That being said I am fine with this: +1. But it should at least pass ci I guess.

Re-triggered the gating, and it's green now.

vuntz:
Got merged in https://github.com/crowbar/barclamp-network/pull/331

PR #264, opened 2016-01-14
neutron: Fix order constraint for neutron-ha-tool

vuntz:
Right now, the order explicitly requires haproxy to be started before
neutron-server, which results in neutron-server being stopped and
started when migration haproxy.

Instead, do not require an order in the dependencies of neutron-ha-tool.
Also depends on postgresql and rabbitmq since without them, this won't
work.

Backport of https://github.com/crowbar/crowbar-openstack/pull/55


rhafer:
+1

aplanas:
+1

PR #265, opened 2016-01-21
Add attribute to disable creation of fixed/floating networks

vuntz:
This is a common request we receive; ideally, we would simply move the
definitions of the networks to the barclamp, and add a way to disable
them there, but in the meantime, this should be good enough.

cc @rhafer 

I will forward-port to crowbar-openstack if there's agreement on this.

### DIFF
--- a/chef/cookbooks/neutron/recipes/server.rb
+++ b/chef/cookbooks/neutron/recipes/server.rb
@@ -344,7 +344,8 @@ end
 # have multiple routers with the same name). To avoid this race-condition we
 # make sure that the post_install_conf recipe is only executed on a single node
 # of the cluster.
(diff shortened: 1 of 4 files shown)


rhafer:
In general I am fine this. I guess it will not be possible to deploy tempest when "create_default_networks"  is false?

vuntz:
> In general I am fine this. I guess it will not be possible to deploy tempest when "create_default_networks" is false?

I guess you're correct. But I don't think there's a way around that?

vuntz:
PR for crowbar-openstack: https://github.com/crowbar/crowbar-openstack/pull/216

dirkmueller:
+1

PR #266, opened 2016-02-18
Fix bridge_mapping for openvswitch setups (bsc#967009)

rhafer:
We're creating the floating network as a "flat" provider network therefore we
need to create the correct bridge_mappings on the network node for it.
(https://bugzilla.suse.com/show_bug.cgi?id=967009)


vuntz:
Is it also something for master?

vuntz:
Assuming this works, +1.

rossella:
+1 , for master no fix should be needed, since we set the bridge_mappings already to support multiple external networks

PR #267, opened 2016-02-26
Fix database attributes of the floating network and its router port (…

rhafer:
…bsc#967009)

Because of several bugs (at least bsc#946882 and bsc#955811) we have setups
where the floating network was created with the wrong provider network
settings. While this somehow continued to work and got unnoticed it will
create problems on systems upgrading from juno to liberty. This fix tries to
to re-adjust the attributes directly neutron database to their correct values
to avoid problems during the upgrade.

https://bugzilla.suse.com/show_bug.cgi?id=967009

### DIFF
--- a/chef/cookbooks/neutron/files/default/crowbar-fix-router-port-binding
+++ b/chef/cookbooks/neutron/files/default/crowbar-fix-router-port-binding
@@ -1,0 +1,44 @@
+#!/usr/bin/python
+
+import sqlalchemy
+import sys
+from oslo_config import cfg
+
+_cli_opts = [
+    cfg.StrOpt('router',
+               required=True,
+               help='name of the network to update'),
+]
+
+_db_opts = [
+    cfg.StrOpt('connection',
+               deprecated_name='sql_connection',
+               default='',
+               secret=True,
+               help='URL to database'),
+]
+
+CONF = cfg.ConfigOpts()
+CONF.register_cli_opts(_db_opts, 'database')
+CONF.register_cli_opts(_cli_opts)
+CONF(project='neutron')
+
+db_uri = CONF.database.connection
+
+router = CONF.router
+
+db = sqlalchemy.create_engine(db_uri)
+try:
+    connection = db.connect()
+except sqlalchemy.exc.SQLAlchemyError as e:
+    print >>sys.stderr, 'Cannot connect to database: %s' % e
+    sys.exit(1)
+
+ret = connection.execute(
+    "UPDATE ml2_port_bindings set vif_type='unbound' "
+    "FROM routers,ports "
+    "WHERE ml2_port_bindings.vif_type='binding_failed' "
+    "AND routers.name=%s "
+    "AND routers.id=ports.device_id "
+    "AND ports.device_owner='network:router_gateway' "
+    "AND ports.id=ml2_port_bindings.port_id;", router)
--- a/chef/cookbooks/neutron/recipes/post_install_conf.rb
+++ b/chef/cookbooks/neutron/recipes/post_install_conf.rb
@@ -92,6 +92,62 @@ when 'ml2'
   else
     Chef::Log.error("default provider network ml2 type driver '#{ml2_type_drivers_default_provider_network}' invalid for creating provider networks")
   end
+
+  # *START* workaround for https://bugzilla.suse.com/show_bug.cgi?id=967009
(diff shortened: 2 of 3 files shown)


chef/cookbooks/neutron/recipes/post_install_conf.rb
```diff
@@ -92,6 +92,49 @@ def mask_to_bits(mask)
...
+    cookbook_file "crowbar-fix-router-port-binding" do
+      source "crowbar-fix-router-port-binding"
+      path "/usr/bin/crowbar-fix-router-port-binding"
+      mode "0755"
+    end
+
+    execute "reset port binding for floating network (bsc#967009)" do
+      command "/usr/bin/crowbar-fix-router-port-binding --router router-floating"
+      action :run
+      only_if "out=$(#{neutron_cmd} router-list); [ $? != 0 ] || echo ${out} | grep -q router-floating"
```
rhafer: I would favor a somewhat more specific `onyl if` condition here.

chef/cookbooks/neutron/recipes/post_install_conf.rb
```diff
@@ -92,6 +92,49 @@ def mask_to_bits(mask)
...
+      "--physnet #{ext_physnet_map["nova_floating"]} --segmentation_id 0 --type flat " \
+      "--tenant_id #{keystone_settings['service_tenant_id']} --network floating"
+    action :run
+    only_if "out=$(#{neutron_cmd} net-show floating -f value -F provider:network_type); " \
+      "[ $? == 0 ] && [ ${out} != \"flat\" ]"
+  end
+
+  file "/usr/bin/crowbar-fix-floating-provider-attributes" do
+    action :delete
+  end
```
vuntz: You're deleting that script, but not the other one.

rhafer: Yeah, because I didn't have a good only_if condition the other one would have been recreated with every chef-client run anyway. But I fixed that now with the flag on the node attributes.

rhafer:
mkcloud ci failed because of a typo. Should be fixed now.

rhafer:
Things I could successfully test:
- upgrading a system deployed with GM (no updates), to verify the fix for the provider attributes 
- upgrading a system deployed with GM+updates, to verify the fix for the router port

Both setups were using ovs+vlan

chef/cookbooks/neutron/files/default/crowbar-fix-router-port-binding
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/python
+
+import sqlalchemy
+import sys
+from oslo_config import cfg
+
+#from neutron import context
+#from neutron.common import utils
+#from neutron.db import api
+#from neutron.plugins.ml2 import db
```
vuntz: I guess you can remove these lines.

chef/cookbooks/neutron/recipes/post_install_conf.rb
```diff
@@ -92,6 +92,62 @@ def mask_to_bits(mask)
...
+      end
+    end
+
+    ruby_block "mark node for having the floating network fixed (bsc#967009)" do
+      block do
+        node[:neutron][:floating_router_fix] = true
+        node.save
+      end
+    end
+  else
```
vuntz: Any reason to do it in a `else`? It should be fine to do at the end of the `unless` block.

vuntz:
Minor comments, +1 overall.

chef/cookbooks/neutron/files/default/crowbar-fix-router-port-binding
rhafer: Oops. Yes. Those are a leftover from a different experiment to fix the problem.

chef/cookbooks/neutron/recipes/post_install_conf.rb
rhafer: On friday I still though it would be cleaner to only remove them when they are for sure node needed any longer, but I guess that doesn't really matter much. I'll change it.

rhafer:
@vuntz I think I addressed the remaining bits.

mjura:
+1

dirkmueller:
+1, although I wouldn't like the temporary scripts being created in /usr/bin.. anyway.

PR #268, opened 2016-03-10
Fix 031_enable_floating_interface migration (bsc#967858)

rhafer:
When runnign the migrations 030_fix_rename_l3_role.rb and
031_enable_floating_interface.rb directly after each other (no chef-client run
on the clients inbetween) the latter migration will fail because the search for
the new "neutron-network" role does not return any nodes. To fix that, search
the network node by the new and the old names of the role.

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=967858


vuntz:
+1

toabctl:
+1

